PR #13, opened 2017-05-29
Comment the world. Re-factor Block.ts into easier to manage packages.

amiller-gh:
- Almost every class and method is commented tsdoc style.
- Pulled all BlockObject classes into their own files under /Block
- Block.ts moved to Block/index.ts which combines individual files under /Block
- Shared methods and functionality for BlockObjects pulled out into /Block/Base.tx
- StateContainer and associated methods pulled out into standalone class to be consumed like PropertyConcerns

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "compile": "rm -rf dist && tsc -p test/tsconfig.json",
     "postinstall": "npm run compile",
     "pretest": "npm run compile && tslint --project tsconfig.json",
-    "prepublishOnly": "npm run compile",
+    "prepublish": "npm run compile",
     "tslint": "tslint --project tsconfig.json",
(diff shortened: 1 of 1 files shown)


src/Block/BlockClass.ts
```diff
@@ -0,0 +1,83 @@
+import selectorParser = require("postcss-selector-parser");
+import { OptionsReader } from "../options";
+import { OutputMode } from "../OutputMode";
+import { CompoundSelector } from "../parseSelector";
+import { isState } from "../BlockParser";
+import { Base, StateContainer } from "./Base";
+import { Block, BlockObject } from "./index";
+
+/**
+ * Represents a Block Class key selector.
```
epicmiller: Oops, copy paste error, sorry

PR #118, opened 2018-04-24
Change the zombie emoji

styfle:
The zombie emoji doesn't look quite right on Windows.
I assume any system that doesn't have the male/female zombie modifiers will also have this issue.

zombie

This makes the zombie gender-neutral so readers can focus on the dead code elimination part 😉


chriseppstein:
Thanks for the patch!

amiller-gh:
Thank you @styfle!

styfle:
Thank you for the wonderful documentation! 🎉 

The landing page with the Code, Compile, Optimize steps is superb 👌

PR #119, opened 2018-04-24
Fix table of contents links

brianchung808:
A few of the links in the toc were broken. I went through and copied the actual link from the section and replaced it in the toc.

emojis in section headers make linking weird


chriseppstein:
merged as https://github.com/linkedin/css-blocks/commit/679f007c9de13dbfe078e84866e976825c4e8d7c

chriseppstein:
Thanks for the patch!!!

Issue #122, opened 2018-04-24
support jsx without typescript


chriseppstein:
Users using js, shouldn't need to have a dependency on the typescript module.

https://github.com/linkedin/css-blocks/blob/master/packages/%40css-blocks/jsx/src/Analyzer/index.ts#L13

jonathantneal:
Input JS file (with `@css-blocks/jsx` installed):

```js
const CssBlocks = require('@css-blocks/jsx');
```

Produces this error:

```
Error: Cannot find module 'typescript'
```

amiller-gh:
Ahh, I think this is a bug with our monorepo dependencies. 

We've pulled the `typescript` dependency out to the root of the monorepo, because for all other projects its only used in dev mode. Unfortunately, until Babel 7 lands, we need to pre-compile typescript files using typescript itself, so `typescript` needs to be listed as a dependency of `@css-blocks/jsx`, which it currently is not.

Quick fix: Add `typescript` back as a dependency to `@css-blocks/jsx`.
Long term fix: Integrate with Babel 7 and enable typescript / flow AST parsing if needed.

amiller-gh:
@jonathantneal to un-block yourself in the short term, you can try to `npm install typescript` in your playyground repo. `@css-blocks/jsx` should be able to discover it then, even if we aren't using it in your build 👍 

I'll work on this issue later tonight – both the short term and long term fix.

chriseppstein:
@amiller-gh lock to an exact version if you bring in any pre-release software 😅

amiller-gh:
Short Term Fix: https://github.com/linkedin/css-blocks/pull/129
Long Term Fix: https://github.com/linkedin/css-blocks/pull/132

@chriseppstein: Pick one 😁

chriseppstein:
Babel 7 👍

PR #129, opened 2018-04-25
fix: Add typescript package as a depenency to @css-block/jsx.

amiller-gh:
Addresses short term fix to #122


amiller-gh:
Yes, there are no lock file changes here – the actual project install stays the same 👍

chriseppstein:
We're going to land #132 instead

PR #132, opened 2018-04-25
feat: Upgrade the JSX integration to use Babel 7 instead of typescript.

amiller-gh:
Addresses long term fix for #122


packages/@css-blocks/website/config/webpack.config.dev.js
```diff
@@ -214,7 +213,13 @@ module.exports = {
                   cacheDirectory: true,
                   compact: true,
                   parserOpts: {
-                    plugins: [ "jsx" ]
```
amiller-gh: I want to replace this babel plugin with a fully fledged loader that does all the babel modifications for you. This will simplify Webpack integration for the end user so they no longer have to think about parser options or integrating with a babel plugin.

I also have a theory that we can consolidate the Plugin and first Analyzer loader. This means our webpack integration story becomes: "Add the Analysis and Rewriter loader to your project where they make sense. If you have type compilation, run Analysis before type compilation and the Rewriter before.".

amiller-gh: But! That is for another PR 🙂

PR #135, opened 2018-04-25
Fix the header image in the README

wodin:


amiller-gh:
@chriseppstein, we'll need a re-deploy of the website after this is pulled in. This will also break again if the asset ever changes without updating this hashed URL too – are we okay with that, or do we want a slightly more robust hosting solution for README assets?

amiller-gh:
@wodin, mind updating in `src/packages/@css-blocks/core/README.md` too? 😄

chriseppstein:
@amiller-gh Yeah, I thought webpack has things for this, I guess we're not webpacking right?

amiller-gh:
No, we just build our typedoc site into our website's public directory right now – no webpacking involved, we treat it as a static asset.

amiller-gh:
But, these image URLs also have to work on sites that scrape READMEs too, so we should use a resolvable URL in source: https://www.npmjs.com/package/@css-blocks/core

wodin:
I've made the same change in packages/@css-blocks/core/README.md
I agree that the changes seem fragile given the hash in the filename, but thought I'd send in the PR and let you decide whether it was a good idea :)

amiller-gh:
Merged 🎉 We'll get a longer term fix in for this later. We'll probably want to add our docs generation flow to the webpack build...

PR #139, opened 2018-04-25
Update link to ElementAnalysis docs

jonathantneal:
> You can read more about these style tracking methods on the [ElementAnalysis API documentation][TODO]

Currently, this part of the documentation is marked as a TODO. If the following link is correct, could it instead point here? http://css-blocks.com/api/classes/_css_blocks_core.elementanalysis.html


amiller-gh:
Thank you @jonathantneal!! 👍

PR #141, opened 2018-04-25
fix: Fix problems with JSX and Webpack integrations.

amiller-gh:
- Parse any imported css file as a block file, not just those ending in .block.css.
 - Rewrite to 'className' attribute, not 'class'.
 - Add '.js' to list of Analyzable file extensions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "@types/chai": "^3.5.2",
     "@types/debug": "0.0.29",
     "@types/loader-utils": "^1.1.3",
+    "@types/minimatch": "^2.0.29",
     "@types/mocha": "^2.2.41",
     "@types/node": "^8.0.0",
     "@types/prettier": "^1.8.0",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
     "@types/chai": "^3.5.2",
     "@types/debug": "0.0.29",
     "@types/loader-utils": "^1.1.3",
-    "@types/minimatch": "^2.0.29",
     "@types/mocha": "^2.2.41",
     "@types/node": "^8.0.0",
     "@types/prettier": "^1.8.0",
--- a/packages/@css-blocks/jsx/package.json
+++ b/packages/@css-blocks/jsx/package.json
@@ -60,6 +60,7 @@
     "babel-types": "7.0.0-beta.3",
     "babylon": "7.0.0-beta.46",
     "debug": "^2.6.8",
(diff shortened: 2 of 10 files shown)


amiller-gh:
> Add '.js' to list of Analyzable file extensions.

This may not be a good thing – it will crawl *ALL* possible js files, not just jsx files, meaning we possibly do a *LOT* of extra work during our compilation. Need some feedback from the JSX community if it is a common pattern to use a `js` extension for `jsx` files in projects that are exclusively `jsx` or if this is an anti-pattern.

matthiaskern:
I think it is quite common to give files containing jsx a `.js` extension. e.g. https://github.com/facebook/create-react-app does so by default in the scaffolded projects

chriseppstein:
@matthiaskern But we need to make sure a project that is using css-modules can migrate to css-blocks. A file that is using CSS Blocks will need to import and call into other react components that are still using css-modules, so if we just crawl, we need a way to discern which imports are css-blocks files and which are css files that aren't.

If a project really wants to use `<blockname>.css` they are basically saying they aren't using style-loader/css-modules. That's not a great assumption for most, especially  if they have re-usable react components via npm.

amiller-gh:
@chriseppstein, do we still want to support mixed-world component files where you could possible import both a block and a css-module? If no, then we can safely decide on if we should parse imported css files as blocks or not based on some kind of template-level identifier instead of block import paths.

We have proposals for an updated JSX syntax which would require consumers to import a css-blocks package in their file, that may be a valid way to identify CSS Blocks ready JSX components.

chriseppstein:
I'm fine adding support for the `.js` extension for components. But we're going to crawl the imports we find, right? We need to make sure our crawler/rewriter strategy is correctly aligned with the loader config. in a project where everything is homogoneous it doesn't cause any problems, but apps that are adopting css-blocks will have an app where some components are css-blocks and some are css-modules.

In terms of a component that is using both css-blocks and css-modules, it's fine, I think we need to provide specific guidance on how to do that successfully in terms of best practices and also by giving errors in some cases. For right now, I'd like to make sure we support applications where components are either 100% css-blocks or not in terms of the css they consume.

I don't think this PR doesn't meets that functionality threshold.

amiller-gh:
The loader would start to run on them but short-circuit when it saw an extension it didn't recognize. Don't think it ever got to the parse phase.

👍 on `*.block.css`

PR #154, opened 2018-04-29
Fixes root state comment on README

fabiomcosta:
My guess is that for `.myClass[state|rootState]` the code would be `styles.myClass.rootState()`, but the code is `styles.rootState()`, so I believe the documentation comment is incorrect and should change as stated by the PR.


amiller-gh:
Good catch! Thank you 👍

PR #157, opened 2018-04-29
Flatten entry array & resolve files without extension

ForsakenHarmony:
might want to move the flatten function somewhere else, please tell me where

### DIFF
--- a/packages/@css-blocks/jsx/src/Analyzer/index.ts
+++ b/packages/@css-blocks/jsx/src/Analyzer/index.ts
@@ -9,7 +9,6 @@ import traverse from "babel-traverse";
 import * as babylon from "babylon";
 import * as debugGenerator from "debug";
 import * as fs from "fs";
-import * as path from "path";
 
 import { CssBlocksJSXOptions } from "../options";
 import { JSXParseError } from "../utils/Errors";
@@ -143,7 +142,7 @@ export class CSSBlocksJSXAnalyzer extends Analyzer<TEMPLATE_TYPE> {
    * @param opts Optional analytics parser options.
    */
   public parseFile(file: string): Promise<JSXAnalysis> {
-    file = path.resolve(this.options.baseDir, file);
+    file = require.resolve(file);
     return new Promise((resolve, reject) => {
       fs.readFile(file, "utf8", (err, data) => {
         if (err) {
--- a/packages/@css-blocks/webpack/src/Plugin.ts
+++ b/packages/@css-blocks/webpack/src/Plugin.ts
@@ -1,5 +1,5 @@
 import { TemplateTypes } from "@opticss/template-api";
-import { ObjectDictionary, objectValues } from "@opticss/util";
+import { ObjectDictionary, objectValues, whatever } from "@opticss/util";
 import * as debugGenerator from "debug";
 import { postcss } from "opticss";
 import * as path from "path";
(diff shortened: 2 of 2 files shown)


packages/@css-blocks/webpack/src/Plugin.ts
```diff
@@ -114,7 +114,10 @@ export class CssBlocksPlugin
       entries = webpackEntry;
     }
     else if (typeof webpackEntry === "object") {
-      entries = objectValues(webpackEntry);
+      function flatten(arr: whatever[]): whatever[] {
```
amiller-gh: Lets at least pull this function out to the root of this file. We could move it in to `@opticss/util`, but I think thats overkill :) We don't have a utils package in the css-blocks repo yet – I think we want to keep it that way if possible.

packages/@css-blocks/jsx/src/Analyzer/index.ts
```diff
@@ -143,7 +142,7 @@ export class CSSBlocksJSXAnalyzer extends Analyzer<TEMPLATE_TYPE> {
    * @param opts Optional analytics parser options.
    */
   public parseFile(file: string): Promise<JSXAnalysis> {
-    file = path.resolve(this.options.baseDir, file);
```
amiller-gh: It looks like this change caused a few failing tests – what was the intent here?

amiller-gh: (Totally possible that its a problem with our test harness 🙂 )

ForsakenHarmony: nah probably caused the failing tests, the problem is it currently doesn't resolve files without extension

packages/@css-blocks/webpack/src/Plugin.ts
ForsakenHarmony: 👌

packages/@css-blocks/webpack/src/Plugin.ts
```diff
@@ -322,3 +322,7 @@ export class CssBlocksPlugin
     this.applyPluginsAsync("block-compilation-complete", result, cb);
   }
 }
+
+function flatten(arr: whatever[]): whatever[] {
+  return arr.reduce((acc, val) => (acc as whatever[]).concat(Array.isArray(val) ? flatten(val) : val), []) as whatever[];
+}
```
chriseppstein: A better signature for this is:

```ts 
function flatten<T extends whatever>(arr: T[] | T[][]): T[];
```

Then no casting will be needed.

packages/@css-blocks/webpack/src/Plugin.ts
```diff
@@ -114,7 +114,7 @@ export class CssBlocksPlugin
       entries = webpackEntry;
     }
     else if (typeof webpackEntry === "object") {
-      entries = objectValues(webpackEntry);
+      entries = flatten(objectValues(webpackEntry)) as string[];
```
chriseppstein: With the signature below this doesn’t need to be cast.

packages/@css-blocks/jsx/src/Analyzer/index.ts
ForsakenHarmony: know of a better way?

packages/@css-blocks/webpack/src/Plugin.ts
ForsakenHarmony: with that I'm getting

chriseppstein: Ok this was a bit tricky. The type of reduce wasn't smart enough to let the type checker work out what was going on. Also, the definition I provided was only two levels of arrays, but the code would handle it recursively. I refactored that function and made it part of our opticss utilities: https://github.com/linkedin/opticss/commit/f104638354ed4c7dedd13b09fead5c1741885cbf

PR #168, opened 2018-05-09
chore: Update all dependencies.

DanielRuf:
This also fixes the Node.js 10 setup.


DanielRuf:
Some tsc fails, what could be the cause? Fix typings or upgrade tsc?

chriseppstein:
I just did this for opticss last night. Thanks for trying to tackle it.

I noticed that @types/webpack is 4.x and webpack is 3.8 so that's probably going to cause some issues.

I'll try running the build after I'm done whacking on opticss and see what the issues are.

DanielRuf:
Probably some tslint misconfiguration. Not yet sure. Ok =)

PR #172, opened 2018-05-16
Improve docs for webpack and jsx

bitttttten:
I made the JSX consistent with the Webpack examples.

Also brought more "ghost"/"undefined" variables into the snippets. I am not sure if this is more confusing but it still follows the consistency of using create-react-app varialbes (i.e. the paths.indexJs).


amiller-gh:
I'm SO sorry this took so long to merge! Thanks for making our docs better 🙂

PR #178, opened 2018-05-24
Adding first acceptance tests to @css-blocks/ember-cli

ef4:
This makes acceptance tests work in the dummy app and adds two simple ones. I was going to add more tests, but I realized I'm unclear on what the right semantics are here. 

1. The current code doesn't seem to be aware of the wrapping element that Ember applies to classic components by default. An example of the wrong behavior is `local-component`, which is getting turned yellow as it should, but the style is being applied at the wrong level (the component's true root element is the parent of the div that appears in its template). Personally, I'd be fine with css-blocks only supporting components that don't have a wrapping div (which is possible today either by enabling the template-only glimmer components optional feature or by setting `tagName` to ""), but if so we should at least try to warn people when they're doing the wrong thing.

2. Fragmentary templates (ones with zero or many top-level elements) are supported and encouraged. For example, the application template in this addon's dummy app is fragmentary (it has no root element). Components are also allowed to be fragmentary (and this will be even easier to do going forward, once components with an automatic wrapping div stop being the default).

It looks to me like there are two things css-blocks can do to work pleasantly with fragmentary components:

 1. Tolerate not having a real root element so long as nobody is trying to apply styles to the `:scope` selector. A template with many top-level elements can accompany a stylesheet with many literal class selectors and the result is unambiguous, and can still benefit from css-block's forced encapsulation and optimizations.

 2. Allow the root element to be designated via syntax within the template. This is probably not a css-blocks framework-level feature, just an addition to the glimmer template analyzer. When a root element is explicitly designated, that would unlock usage of the `:scope` selector. A syntax for designating the root element is something we already plan to need in glimmer components for multiple other purposes  (like designating where to put HTML attributes or event handlers that callers try to apply to your component).

### DIFF
--- a/packages/@css-blocks/ember-cli/package.json
+++ b/packages/@css-blocks/ember-cli/package.json
@@ -16,7 +16,7 @@
     "build": "ember build",
     "lint:js": "eslint ./*.js addon addon-test-support app blueprints config lib server test-support tests",
     "start": "ember serve",
-    "test": "ember test",
+    "test": "EMBER_CLI_MODULE_UNIFICATION=true ember test",
     "test:all": "ember try:each"
   },
   "dependencies": {
--- a/packages/@css-blocks/ember-cli/package.json
+++ b/packages/@css-blocks/ember-cli/package.json
@@ -44,6 +44,7 @@
     "ember-export-application-global": "^2.0.0",
     "ember-load-initializers": "^1.1.0",
     "ember-maybe-import-regenerator": "^0.1.6",
+    "ember-native-dom-helpers": "^0.6.2",
     "ember-resolver": "^4.0.0",
     "ember-source": "https://s3.amazonaws.com/builds.emberjs.com/canary/shas/5431f71df060e7b5c82c05c300f40e9151d135e8.tgz",
     "ember-source-channel-url": "^1.0.1",
--- a/packages/@css-blocks/ember-cli/tests/acceptance/styles-applied-test.js
+++ b/packages/@css-blocks/ember-cli/tests/acceptance/styles-applied-test.js
@@ -1,0 +1,18 @@
+import { module, test } from 'qunit';
+import { visit } from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+import { find } from 'ember-native-dom-helpers';
+
+module('Acceptance | styles applied', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('route stylesheet is applied', async function(assert) {
+    await visit('/posts/new');
+    assert.equal(getComputedStyle(find('[data-test="posts-new-h3"]')).color, 'rgb(0, 128, 0)');
+  });
+
+  test('block reference style is applied', async function(assert) {
+    await visit('/posts/new');
+    assert.equal(getComputedStyle(find('[data-test="posts-new-h3"]'))['font-style'], 'italic');
+  });
+});
--- a/packages/@css-blocks/ember-cli/tests/dummy/config/environment.js
+++ b/packages/@css-blocks/ember-cli/tests/dummy/config/environment.js
@@ -48,6 +48,7 @@ module.exports = function(environment) {
     ENV.APP.LOG_VIEW_LOOKUPS = false;
 
     ENV.APP.rootElement = '#ember-testing';
+    ENV.APP.autoboot = false;
   }
 
   if (environment === 'production') {
--- a/packages/@css-blocks/ember-cli/tests/dummy/src/ui/routes/posts/new/template.hbs
+++ b/packages/@css-blocks/ember-cli/tests/dummy/src/ui/routes/posts/new/template.hbs
@@ -2,5 +2,6 @@
   class="typography"
   state:typography.bold
   state:typography.italic
+  data-test="posts-new-h3"
 >Posts – New</h3>
 {{outlet}}
--- a/yarn.lock
+++ b/yarn.lock
@@ -5160,6 +5160,13 @@ ember-maybe-import-regenerator@^0.1.6:
     ember-cli-babel "^6.0.0-beta.4"
     regenerator-runtime "^0.9.5"
 
+ember-native-dom-helpers@^0.6.2:
+  version "0.6.2"
+  resolved "https://registry.npmjs.org/ember-native-dom-helpers/-/ember-native-dom-helpers-0.6.2.tgz#ad1f82d64ac9abdd612022f4f390bdb6653b3d39"
+  dependencies:
+    broccoli-funnel "^1.1.0"
+    ember-cli-babel "^6.6.0"
+
 ember-qunit@^3.3.2:
   version "3.4.0"
   resolved "https://registry.npmjs.org/ember-qunit/-/ember-qunit-3.4.0.tgz#47a60c2b889cd4b4a46380bf9da2b10115c0eae7"


amiller-gh:
> 1. The current code doesn't seem to be aware of the wrapping element that Ember applies to classic components by default. An example of the wrong behavior is local-component, which is getting turned yellow as it should, but the style is being applied at the wrong level (the component's true root element is the parent of the div that appears in its template). Personally, I'd be fine with css-blocks only supporting components that don't have a wrapping div (which is possible today either by enabling the template-only glimmer components optional feature or by setting tagName to ""), but if so we should at least try to warn people when they're doing the wrong thing.

Agreed! Is there a good way to catch this on a per-component basis at build time? I'm all for only supporting CSS Blocks only supporting non-wrapped components, but wasn't aware of a good way to lint for this.

Alternatively, we can ignore that this wrapper element exists – descendent selectors aren't allowed across component boundaries (caveat: global states, but I don't think that should pose a problem here), so I don't believe the additional un-expected element should cause any styling bugs – pinging @chriseppstein to confirm.

Ignoring the element for styling purposes *may* cause no bugs, but does requires a little extra mental gymnastics to reason about your styles. However, this mental overhead may go away if we adopt another method for referencing blocks, described below.

> 2. Fragmentary templates (ones with zero or many top-level elements) are supported and encouraged. For example, the application template in this addon's dummy app is fragmentary (it has no root element). Components are also allowed to be fragmentary (and this will be even easier to do going forward, once components with an automatic wrapping div stop being the default).
>
> It looks to me like there are two things css-blocks can do to work pleasantly with fragmentary components:
>
> 1. Tolerate not having a real root element so long as nobody is trying to apply styles to the :scope selector. A template with many top-level elements can accompany a stylesheet with many literal class selectors and the result is unambiguous, and can still benefit from css-block's forced encapsulation and optimizations.
>
> 2. Allow the root element to be designated via syntax within the template. This is probably not a css-blocks framework-level feature, just an addition to the glimmer template analyzer. When a root element is explicitly designated, that would unlock usage of the :scope selector. A syntax for designating the root element is something we already plan to need in glimmer components for multiple other purposes (like designating where to put HTML attributes or event handlers that callers try to apply to your component).

I'm not a fan of option 1, it feels like it adds yet another potential pit fall for consumers to fall in to, adding to their frustration with the framework. 

Option 2 could be wonderful, especially if it become a sanctioned pattern for Glimmer. I'm interested in exploring that more!

Or a 3rd option: Chris and I toyed with an explicit import syntax for Block files in Glimmer that may make this problem go away entirely. I'd be curious if you feel like this goes against the  "Ember Way", or if you have any other opinions or syntax ideas.

With the current Ember + CSS Blocks proposal, every component or route template may have an associated Block file. This is assigned to it though convention (presumably in non-module-unified and non-pods apps the `/styles` directory would mimic the structure of `/templates`?) If a component wants to reference another block, they add an explicit `@block-reference` and use the block in their templates under the aliased name (see [`src/ui/routes/posts/new/stylesheet.css`](https://github.com/linkedin/css-blocks/pull/174/files#diff-1b726646148b17b408c3873715067255) in the PR)

In the JSX implementation, `@block-reference` calls in the Block file are meant to be benign, and don't actually inject any extra data into the template scope – to use multiple Blocks in a JSX file you need to explicitly `import localName from "path/to/block.css";` in your javascript for each Block used. (see code examples on [css-blocks.com](www.css-blocks.com) This has the nice syntactic side effect of forcing users to explicitly apply the `:scope` styles to an element in their template.

We could change the API of Glimmer/Ember templates to require explicit imports in the template. For example:
```css
/* path/to/block1.css */
:scope { /* ... */ }
:scope[state|foobar] { /* ... */ }
.klass1 { /* ... */ }
.klass1[state|bizbaz] { /* ... */ }
```
```css
/* path/to/block2.css */
:scope { /* ... */ }
.klass2 { /* ... */ }
```
```hbs
{{block-import "block1" "path/to/block1.css"}}
{{block-import "block2" "path/to/block2.css"}}

<div class="block1" state:foobar={{isFooBar}}>
{{!-- Or maybe: `<div scope:block1>`? --}}
  <div 
      class="block1.klass1" 
      state:block1.bizbaz={{isBizBaz}}
  ></div>
</div>

<div class="block2">
  <div class="block2.klass2"></div>
</div>
```
At build time, all fake `block-import` helpers in the file (could be a pragma comment?) would be stripped and the classes/states used would be re-written exactly as they are now. In addition to better supporting multiple roots in a template, removing the possibly unexpected (and currently glimmer-only) scope injection of `@block-reference`, and the magic application of Block scopes, this also has the added benefit of almost fully aligning the Glimmer and JSX syntaxes.

Its a thought 🙂 

---

More to the point here though – PR looks great 👍Looks like the tests aren't configured to run with the module unification env variable set though. Once thats fixed and the Travis tests pass I'll be happy to merge in!

ef4:
> I'm not a fan of option 1, it feels like it adds yet another potential pit fall for consumers to fall in to, adding to their frustration with the framework.

The pitfall exists already, in that templates just fundamentally don't represent a single-element in glimmer or ember. If blocks are going to stay strictly to that semantic, there will remain a gap that forces you to write unidiomatic glimmer templates if you want to establish a clear one-to-one conventional relationship between css blocks and glimmer templates (which I strongly want to establish).

I threw out the suggestion (of supporting blocks without a real root Element as long as no styles are targeting `:scope`) partly to hear if it's even feasible given CSS blocks internals. Intuitively it seems not a big deal?

I can understand that adding any kind of additional "mode" to an api design is not a cost to incur lightly. That is totally an RFC-level API design question, not to be jumped into on a whim, and my intuitions could be wrong. 

But my intuition is that thinking of a stylesheet and a template as their own independent little world where you can (1) start writing some basic CSS and HTML and they will Just Work by matching up with each other (while never leaking outside that world) for all the easy cases and (2) where they don't work, a compiler will be able to tell you exactly why and how to address it. I'm a big fan of the model CSS Blocks has been taking precisely because it is so learnable and embraces so much existing knowledge of CSS while having a compiler that can guide people into learning the things that are new.

> Option 2 could be wonderful, especially if it become a sanctioned pattern for Glimmer. I'd definitely be interested in exploring that more!

I agree but keep in mind this doesn't really address the semantic mismatch, because that syntax exists to support the _distinction_ between components that have a root element and ones that don't. Both cases are idiomatic.

> Or a 3rd option: Chris and I toyed with an explicit import syntax for Block files in Glimmer that may make this problem go away entirely. I'd be curious if you feel like this goes against the "Ember Way", or if you have any other opinions / syntax ideas.

Yeah, strong negative on that. It's added line noise for solving a problem we can readily solve in the compilers instead (speaking of both glimmer's and css block's). In a big app with a big team, you will want to have standards for what is the right place to put the block file that goes with a given component, and as soon as you've standardized that, forcing devs to repeat it everywhere by hand (and risk getting it wrong for no benefit) is just a net drain overall on increased bikeshedding / nitpicking.

This is a case where it doesn't really "make the problem go away", it just punts the problem off onto every team to solve independently. Now, it's totally appropriate to _not_ solve this class of problem in CSS blocks itself, but integrators should be able to. Which is why I zoomed in on this particular question of supporting optional looser block semantics, since that is really the only core piece that would be needed before I could make an integration do the things I'd want it to do. (I mean, I can probably _already_ make it do what I want, but less hacks is better.)

chriseppstein:
> (like designating where to put HTML attributes or event handlers that callers try to apply to your component)

I have feelings about this that are, perhaps, unrelated to this PR discussion, but might be... because those feelings have informed the opinions that CSS Blocks enforces at the API boundary of components for styling:

I just don't think the caller should dictate a concrete implementation to a component without interfacing through a public API. For styles this means we make it nigh impossible to write a selector that overrides a CSS Blocks-styled element using combinators involving elements unrelated to that block (Exception: global states which is a different kind of public api). Instead we will provide a way to expose a public API for styling a component in well-defined, efficiently-statically-analyzable ways.

_Aside: This means we need a way to declare one or more references to blocks as arguments that can be passed to the component at runtime -- this seems likely to lead to some need to analyze/rewrite the component file, but I can't say for sure._

I see the need to subscribe to events from a component, but I question the decision to let the caller do so at the DOM level directly. I think the component should always proxy that setup to insulate the caller from internal machinations over time. But even better, I'd rather that code subscribe to component-specific events that are decoupled from DOM events entirely.

I've heard that accessibility is a primary use case for this because it's often an oversight by even popular ember components and so it would limit the ability for people to specify important aria attributes on the component's scope element. I would argue that a narrow cutout or use-case specific abstraction for such things is a better path.

> > Or a 3rd option: Chris and I toyed with an explicit import syntax for Block files in Glimmer that may make this problem go away entirely. I'd be curious if you feel like this goes against the "Ember Way", or if you have any other opinions / syntax ideas.

> Yeah, strong negative on that. It's added line noise for solving a problem we can readily solve in the compilers instead (speaking of both glimmer's and css block's). In a big app with a big team, you will want to have standards for what is the right place to put the block file that goes with a given component.

I agree that we want to have a default block file per component and to infer it presence based on conventions. I don't think @amiller-gh was suggesting that we do away with that concept. Our preference is to de-couple `@block-reference` within a block file associated to a template from "make a block available to this template with the following lexically scoped name". We think a good solution is to instead add a statically resolved (and removed) helper invocation (or pragma comment) to templates to take over that concern, but are open to other suggestions if they make more sense in the context of ember's conventions and idioms. Even when there is a default block with a singly-defined scope element, there are use cases for putting the scope class on a different element. I can get on board with automatically injecting the scope class (if defined for that component) onto the component's main element, as long as there is a way to override that behavior. And I think it's reasonable to give an error if there's not a single element that wraps the component and there are selectors that have `:scope` in the key selector within the components' block file.



```hbs
{{block-import "font" from "mydesignsystem/fonts.block.css"}}

<div class="font" state:font.family="sans" state:font.size={{style-if embiggenMode "larger" "normal"}}>
   {{!-- If there was some way to explicitly declare the ember root element that gets the canonical html
           attributes that are passed to the component css blocks could infer the :scope styles from that --}}
   <ember:root ember:tagName="p" ember:apply-attributes="aria-*" state:open={{isOpen}}>
      <!-- Stuff inside -->
   </ember:root>
</div>
```

Whether or not there's a way to say which element is the root element, we still need a way to apply styles for the block to a different element. One such idea is to reserve a single class name like `root` when there are selectors that target the :scope element. But this means that a block cannot have a class with the name of `root` which I don't like. 

```hbs
{{block-import "font" from "mydesignsystem/fonts.block.css"}}

<div class="font" state:font.family="sans" state:font.size={{style-if embiggenMode "larger" "normal"}}>
   <div class="root" state:open={{isOpen}}>
      <!-- Stuff inside -->
   </div>
</div>
```

Another approach would be to expose an expression with special significance. Probably a synthetic helper of some sort:

```hbs
{{block-import "font" from "mydesignsystem/fonts.block.css"}}

<div class="font" state:font.family="sans" state:font.size={{style-if embiggenMode "larger" "normal"}}>
   <div class={{style-root}} state:open={{isOpen}}>
      <!-- Stuff inside -->
   </div>
</div>
```

A third option is to let the template author bind a local name to access the default block where it can't be inferred otherwise.

```hbs
{{block-import "font" from "mydesignsystem/fonts.block.css"}}
{{block-import "defaults" from self}}

<div class="font" state:font.family="sans" state:font.size={{style-if embiggenMode "larger" "normal"}}>
  {{!-- the `defaults` reference in the open state attribute would be optional here --}}
   <div class="defaults" state:defaults.open={{isOpen}}>
      <!-- Stuff inside -->
   </div>
</div>
```

> This is a case where it doesn't really "make the problem go away", it just punts the problem off onto every team to solve independently. Now, it's totally appropriate to not solve this class of problem in CSS blocks itself, but integrators should be able to.

No, I definitely want block reference paths whether in block stylesheets or in templates to fully integrate with ember resolver system and conventions.

>  Intuitively it seems not a big deal?

Using a block as a namespace over simple classes is a reasonable and valid scenario that I want to support. I would never want to require a `:scope` element and a block's classes purposefully do not combinate with `:scope` for this reason. But when styling behaviors arise between classes that's going to always prompt those styles to be factored into a block with a scope element and combinators to other classes/states. that doesn't mean that block has to be the default block for a component -- in general we should expect component templates to use many blocks, even small ones intended specifically for that one component and not shared.

> I'm a big fan of the model CSS Blocks has been taking precisely because it is so learnable and embraces so much existing knowledge of CSS while having a compiler that can guide people into learning the things that are new.

Gosh this makes me happy, because it's exactly what we've wanted to accomplish.

I hope I've clarified some things here?

ef4:
Thanks @chriseppstein, that does clarify things and I'm glad we're pretty much on the same page.

> Using a block as a namespace over simple classes is a reasonable and valid scenario that I want to support. 

Great, I think this case is going to aid adoption, and by itself it's enough to resolve the broken cases we encountered immediately in this addon's dummy app. I agree that as soon as people's usage gets more sophisticated they're going to want to factor out more blocks.

For designating the style root, your examples give me plenty of room to do an experimental implementation so we can get some usage feedback. I'm going to look at the "synthetic helper" option first, because I agree about not making a string like `"root"` be magically special, and because it's very normal for an ember addon to provide a helper.

amiller-gh:
Closing in favor of https://github.com/linkedin/css-blocks/pull/185 🙏 Great conversation here though.

PR #185, opened 2018-07-11
[WIP] Ember cli classic

amiller-gh:


packages/@css-blocks/broccoli/src/index.ts
```diff
@@ -48,11 +58,18 @@ class BroccoliCSSBlocks extends BroccoliPlugin {
...
     // This build step is *mostly* just a pass-through of all files!
     // QUESTION: Tom, is there a better way to do this in Broccoli?
     let files = await readdir(this.inputPaths[0]);
     for (let file of files) {
       file = path.relative(this.inputPaths[0], file);
+
+      // If we're in Classic or Pods mode, every hbs file is an entry point.
+      if (discover && path.extname(file) === ".hbs") { this.entry.push(file); }
+
       await fs.ensureDir(path.join(this.outputPath, path.dirname(file)));
```
stefanpenner: let transition to sync FS functions.

* in general they are more performant (no need to schedule to the event loop and come back), and as we are processing serially through a build. Rarely (if every) does the non-blocking aspect help.
* typically sync errors have better stack traces (although maybe newer node's fixed this)

packages/@css-blocks/ember-cli/addon/helpers/tmp.js
```diff
@@ -0,0 +1,124 @@
+
+const e = (m) => { throw new Error(m); };
```
stefanpenner: this module looks pretty complex, do we have associated unit tests?

packages/@css-blocks/glimmer/src/Analyzer.ts
```diff
@@ -66,66 +63,61 @@ export class GlimmerAnalyzer extends Analyzer<TEMPLATE_TYPE> {
...
-      componentDeps.components.forEach(c => components.add(c));
+    this.debug(`Analyzing all templates starting with: ${componentNames}`);
+
+    componentNames.forEach(componentName => {
+      components.add(componentName);
+      try {
+        let componentDeps = this.resolver.recursiveDependenciesForTemplate(componentName);
+        componentDeps.forEach(c => components.add(c));
+      } catch(e){
+        this.debug(`Warning: Could not discover recursive dependencies for component ${componentName}`);
```
stefanpenner: If we do intend to catch here, we should likely only catch errors we expect. For example, we should likely not catch TypeErrors and similar.

That way, we do not hide real issues.

QQ: do we intend to catch here going forward?

packages/@css-blocks/glimmer/src/Analyzer.ts
```diff
@@ -66,66 +63,61 @@ export class GlimmerAnalyzer extends Analyzer<TEMPLATE_TYPE> {
...
+      }
     });
 
     this.debug(`Analyzing all components: ${[...components].join(", ")}`);
 
     components.forEach(dep => {
-      analysisPromises.push(this.analyzeTemplate(dep, depAnalyzer));
+      analysisPromises.push(this.analyzeTemplate(dep));
```
stefanpenner: `components.map` ?

packages/@css-blocks/glimmer/src/Resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+      });
+    }
+  }
+
+  // TODO: We need to automatically discover the file ending here – its not guaranteed to be a css file.
+  private async tmplPathToStylesheetPath(template: string): Promise<string | undefined> {
+    // First try Classic Ember structure.
+    let classic = template.replace('templates/', 'styles/').replace('.hbs', '.block.css');
+    classic = path.join(this.projectDir, this.srcDir, classic);
+    if (await fs.pathExists(classic)) {
```
stefanpenner: `existsSync`

packages/@css-blocks/glimmer/src/Resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+      DEBUG(`Discovered classic Block for template ${template}:`);
+      DEBUG(`  - ${classic}`);
+      return classic;
+    }
+    let pods = path.parse(template);
+    pods.base = "stylesheet.block.css";
+    let podsPath =  path.join(this.projectDir, this.srcDir, path.format(pods));
+    if (await fs.pathExists(podsPath)) {
```
stefanpenner: `existsSync`

packages/@css-blocks/glimmer/src/Resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Given a template identifier, resolve the Block file associated
+   * with the template, if any.
+  */
+  async stylesheetFor(identifier: string): Promise<ResolvedFile | undefined> {
+    if (!this.depAnalyzer) {
+      let stylesheet = await this.tmplPathToStylesheetPath(identifier);
+      if (!stylesheet ) { return undefined; }
+      return new ResolvedFile(
+        (await fs.readFile(stylesheet)).toString(),
```
stefanpenner: `readFileSync(stylesheet, 'UTF8')`

packages/@css-blocks/glimmer/src/Resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+        stylesheet
+      );
+    }
+
+    // TODO: We need to automatically discover the file ending here – its not guaranteed to be a css file.
+    identifier = this.depAnalyzer.project.resolver.identify(`stylesheet:${identifier}`);
+    let file: string = this.depAnalyzer.project.resolver.resolve(identifier);
+    file = path.join(this.projectDir, this.srcDir, file);
+
+    let content = (await fs.readFile(file)).toString();
```
stefanpenner: ^

packages/@css-blocks/glimmer/src/Resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Given a template identifier, resolve the actual template file associated
+   * with it, if any.
+  */
+  async templateFor(identifier: string): Promise<ResolvedFile | undefined> {
+    if (!this.depAnalyzer) {
+      let template = path.join(this.projectDir, this.srcDir, identifier);
+      return new ResolvedFile(
+        (await fs.readFile(template)).toString(),
```
stefanpenner: ^

packages/@css-blocks/glimmer/src/Resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+        identifier,
+        identifier
+      );
+    }
+    try {
+      let template = this.depAnalyzer.project.templateFor(identifier);
+      return new ResolvedFile(template.string, template.specifier, template.path);
+    } catch(e) {
+      return undefined;
```
stefanpenner: We should likely only catch errors we expect, re-throwing exceptional errors. This prevents hiding bugs.

packages/@css-blocks/broccoli/src/index.ts
amiller-gh: 👍

packages/@css-blocks/ember-cli/addon/helpers/tmp.js
amiller-gh: Copy pasta from the unit tested version in `@css-blocks/glimmer` because Ember isn't resolving at build time – issue is commented in the helpers file. This will have to be resolved before it can be merged.

packages/@css-blocks/glimmer/src/Analyzer.ts
amiller-gh: This catch can likely be removed since I've refactored a bit more, will look into it.

packages/@css-blocks/glimmer/src/Resolver.ts
amiller-gh: 👍

packages/@css-blocks/glimmer/src/Analyzer.ts
amiller-gh: Map with side effects makes me squirm...but okay. Will look like: 

```js
[...components]
  .map(this.analyzeTemplate, this)
  .map(Array.prototype.push, analysisPromises);
```

stefanpenner: > Map with side effects makes me squirm

Seems ok then.

amiller-gh:
Addon Status Update: The addon is now a good broccoli citizen and does not reach outside of its tree at at build time 🎉

I'm now focusing on making existing tests pass again and getting addon test coverage. Part of that will be to verify that deeply nested addons don't break the build.

Once this is finished, I think the PR can be pulled in to master a partial ember-cli addon implementation! More iterative work must then be done to:
  - Improve Broccoli caching
  - Add node_modules resolution of `@block-reference`s
  - Update Analyzer APIs to prepare for a Language Server world – this will get rid of the insane Transport objects in the ember-cli and webpack build integrations. 

I'll be writing RFCs for these changes after the addon lands in master.

packages/@css-blocks/ember-cli/tests/integration/template-discovery-test.js
```diff
@@ -0,0 +1,63 @@
...
+    assert.ok(varIsPresent("#link-to-helper-active", "link-to-helper-active"), "Link-to helpers receive active states");
+    assert.ok(varIsPresent("#input-helper", "input-helper"), "Input helpers receive classes");
+    assert.ok(varIsPresent("#textarea-helper", "textarea-helper"), "Textarea helpers receive classes");
+  });
+
+  test("App Route Block Integration", async function(assert) {
+    await visit("/route-block");
+    assert.equal(currentURL(), "/route-block", "Navigated to test case");
+    assert.ok(varIsPresent("#scope", "route-block-scope"), "Scope style applied to root element");
+    assert.ok(varIsPresent("#sub-class", "route-block-class"), "Sub-classes applied to children");
```
chriseppstein: I think it can follow in a separate PR, but I think we need to design some test helpers for ember unit tests that allow the developer to make assertions about block styles. Something like:

```js
import { testHelpers as blockAssertions } from "@css-blocks/ember-cli";
async function (assert) {
  let block = blockAssertions("path/to/file1.block.css");
  let blockScopeElement = document.getElementById("block1");
  let blockClassElement = document.getElementById("block1child");
  block.assertCurrentStyle(blockScopeElement, ":scope");
  block.assertCurrentStyle(blockClassElement, ".aClass[state|foo][state|bar]");
}
```

amiller-gh: Yeah I was thinking the same thing 👍

amiller-gh:
@stefanpenner, @chriseppstein, I think this is ready for another review! At your leisure 🙂 

---

For newcomers just discovering this PR, here are the caveats for this proof-of-concept ember-cli plugin: 

 1. Eager engines are a little hacky, but otherwise work fine. We'll want to make some changes to engines proper to make its behavior more consistent with regular addons, but I think we need to commit the hack for now.
 2. Lazy engines are not possible at this time and will require non-trivial work in in `ember-engines`. I have a failing (skipped) test for them though.
 3. The `{{link-to}}` helper (and possibly other built-in form helpers) will require some special casing in `@css-blocks/glimmer` to have a natural feeling integration. I have failing (skipped) tests in the project.
 4. The optimizer can not be enabled at this time.
 5. The broccoli plugins do not have any kind of caching strategy right now! That, and build impact stats / tests, are coming soon.
 6. `node_module` resolutions for `@block-reference`s do not work yet.

All six of the above items will need to be finished in order to call this addon "done". I'll be tracking these all in separate tickets. 

But otherwise, this works for unlocking CSS Blocks syntax in your addons and apps 🎉

packages/@css-blocks/broccoli/src/index.ts
tomdale: Doesn't this foreclose on the ability to run multiple Broccoli plugins concurrently in the future?

amiller-gh: I don't believe so. We're already running multiple instances of this broccoli plugin (one for each addon, engine, or app). Broccoli (according to spenner) runs the build step synchronously already, so we don't actually gain any perf benefits from async functions and instead are just left with that extra throwaway promise being created for each file system call.

If Broccoli begins running certain build steps asynchronously, this *may* be a perf bottleneck, but we'll need to update synchronous fs calls across the entire ecosystem to un-block it. 

From a CSS Blocks perspective, there is no harm in sync vs async.

NullVoxPopuli:
does this enable support for ember's module unification?
I'd love to get css-blocks going in https://emberclear.io :)

PR #188, opened 2018-08-11
Ember cli classic – merge take 2

amiller-gh:
I screwed up pulling in https://github.com/linkedin/css-blocks/pull/185... it was pointing at a branch that wasn't master ¯\_(ツ)_/¯ 

Lets do this again.


bartocc:
Now that it is merged in master, can we use css-blocks in an ember.js app ?

amiller-gh:
Because its still pre-1.0 feel free to use it, but at your own risk 🙂Would definitely appreciate some help kicking the wheels a bit and seeing what breaks.

Give the addon's readme a read and see if, even with the caveats, it will work for your use cases right now. I expect most of those six outstanding items to be finished in the next few months. I'll have RFC issues for all six up this week. 

https://github.com/linkedin/css-blocks/blob/fb7a54076270bdb4102b2931374ff12803f1d348/packages/%40css-blocks/ember-cli/README.md

Its also not published yet! Waiting on @chriseppstein to cut a beta version of the addon.

.travis.yml
```diff
@@ -6,7 +6,10 @@ language: node_js
 
 node_js:
   - "8"
+<<<<<<< HEAD
```
dbbk: Looks like this git conflict accidentally made it in!

amiller-gh: 😫Lets pretend this never happened...

Issue #192, opened 2018-09-05
Extended `@block` Reference Syntax


amiller-gh:
# Summary
Implementation of this proposal will enable the following features:

 - [x] Rename `@block-reference` to `@block`
 - [x] Additions to the `@block` import syntax
   - [x] Default `@block` imports from a Block module
   - [x] Named `@block` imports from a Block module
   - [x] Default Block exports from a module
   - [x] Named Block exports from a module
 - [x] `default` as a reserved name
 - [x] `css-blocks` package.json config field
     - [x] Custom Block module "main" path
 - [x] CSS Blocks module resolution algorithm
     - [x] Relative Block import path resolution
     - [x] Absolute Block import path resolution
     - [x] `node_module` Block import path resolution

# Detailed design
## Rename `@block-reference` to `@block`
After some time out in the wild, we've received feedback that typing `@block-reference` is onerous and looks weird. 

This proposal would update `@block-reference`:
```css
@block-reference foobar from "path/to/file.block.css";
```
To `@block`:
```css
@block foobar from "path/to/file.block.css";
```
This has the added benefit of providing the option of backwards compatibility / deprecation for this feature change, although we are pre-1.0 and this is not required.

## `@block` Imports
To enable a more robust module delivery system in css-blocks, this proposal would define the API options for `@block` as follows.

| Token | | Definition |
|---|---|---|
| `<block-import>` | ::= | " **@block** " `<blocks-list>` " **from** " `<block-path>` |
| `<blocks-list>`  | ::= | `<default-block>` \| `<named-blocks>` \| `<default-block>` " **,** " `<named-blocks>`|
| `<default-block>` | ::= | `<ident>` |
| `<named-blocks>` | ::= | " **(** " `<named-ref>` { " **,** " `<named-ref>` } " **)** " |
| `<named-ref>` | ::= | `<local-name>` \| `<aliased-block>` |
| `<aliased-block>` | ::= | `<ident>` " **as** " `<local-name>` |
| `<local-name>` | ::= | `<ident>` (Note: Lexer restriction – cannot be "**default**") |
| `<block-path>` | ::= | ' **"** ' `<any-value>` ' **"** ' \| " **'** " `<any-value>` " **'** " |

Examples of this grammars use are 

### Default `@block` imports from a Block module
```css
@block local-name from "path/to/block.css";
```
In the above example, the default exported block from the block file discovered at `<block-path>` will be available under the local name `local-name`.

### Named `@block` imports from a Block module
[...]

PR #196, opened 2018-09-10
feat: Fully Qualified Block Path Node Modules Importer. Issue #192.

amiller-gh:
Executes on module resolution plan from #192.
Companion PR to #210.

 - Consolidate importer logic
 - Clean up promises use (async/await ftw)
 - Enable both fully qualified Block path node_modules resolution
 - Enable default package export discovery for node_modules Block resolution.


chriseppstein:
Merged as 5a9068e6a80160ac2ffcc7672383b669a07debd1

packages/@css-blocks/core/src/importing/NodeJsImporter.ts
```diff
@@ -0,0 +1,136 @@
...
+    from = from ? this.filesystemPath(from, config) : from;
+    let fromDir = from ? path.dirname(from) : config.rootDir;
+    let resolvedPath = path.resolve(fromDir, importPath);
+    if (existsSync(resolvedPath)) { return resolvedPath; }
+    debug(`No relative or absolute Block file discovered for ${importPath}.`);
+
+    // If not a real file, attempt to resolve to an aliased path instead.
+    let alias = this.aliases.find(a => importPath.startsWith(a.alias + path.sep));
+    if (alias) {
+      return path.resolve(alias.path, importPath.substring(alias.alias.length + 1));
```
stefanpenner: I know this has already landed, so I can gladly open a new issue. But could we explore mimicing node's module caching semantics for these resolves?

This will help prevent css-blocks from running into similar problems that eyeglass hit: https://github.com/sass-eyeglass/eyeglass/pull/189

Let me know if i should open a new issue.

stefanpenner: [I've opened an issue to facilitate the conversation](https://github.com/linkedin/css-blocks/issues/216)

PR #197, opened 2018-09-10
fix(broccoli): Rebuild diffs input changes and output patch separately.

amiller-gh:
Fixes a bug where changes are occasionally not picked up.

### DIFF
--- a/packages/@css-blocks/ember-cli/tests/dummy/app/styles/node-modules-resolution.block.css
+++ b/packages/@css-blocks/ember-cli/tests/dummy/app/styles/node-modules-resolution.block.css
@@ -1,5 +1,0 @@
-/* @block-reference other from "in-repo-engine/styles/components/addon-component.block.css"; */
-
-:scope {
-  /* extends: other; */
-}
--- a/packages/@css-blocks/ember-cli/tests/dummy/app/styles/node-modules-resolution.block.css
+++ b/packages/@css-blocks/ember-cli/tests/dummy/app/styles/node-modules-resolution.block.css
@@ -1,5 +1,5 @@
-@block-reference other from "in-repo-engine/styles/components/addon-component.block.css";
+/* @block-reference other from "in-repo-engine/styles/components/addon-component.block.css"; */
 
 :scope {
-  extends: other;
+  /* extends: other; */
 }
--- a/packages/@css-blocks/ember-cli/tests/dummy/lib/in-repo-addon/package.json
+++ b/packages/@css-blocks/ember-cli/tests/dummy/lib/in-repo-addon/package.json
@@ -5,7 +5,7 @@
     "ember-addon"
   ],
   "dependencies": {
-    "@css-blocks/ember-cli": "*",
+    "@css-blocks/ember-cli": "^0.20.0-beta.0",
     "ember-cli": "*",
     "ember-cli-htmlbars": "*",
     "ember-cli-babel": "*"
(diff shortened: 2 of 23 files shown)


packages/@css-blocks/glimmer/test/recursive-template-dependencies-test.ts
```diff
@@ -1,5 +1,6 @@
+import Analyzer from "@amiller-gh/glimmer-analyzer";
```
amiller-gh: Fixing an oversight from a previous commit. Was using an old version of glimmer-analyzer in this test.

packages/@css-blocks/ember-cli/tests/dummy/app/styles/node-modules-resolution.block.css
```diff
@@ -1,5 +1,5 @@
-@block-reference other from "in-repo-engine/styles/components/addon-component.block.css";
+/* @block-reference other from "in-repo-engine/styles/components/addon-component.block.css"; */
```
chriseppstein: Why is this code commented out instead of just marking the test case as skipped?

packages/@css-blocks/ember-cli/tests/dummy/lib/in-repo-addon/package.json
```diff
@@ -5,7 +5,7 @@
     "ember-addon"
   ],
   "dependencies": {
-    "@css-blocks/ember-cli": "*",
+    "@css-blocks/ember-cli": "^0.20.0-beta.0",
```
chriseppstein: This looks like it will be annoying to maintain. why doesn't `*` work?

chriseppstein: (Ditto for the other `package.json` files where there change was made.)

packages/@css-blocks/glimmer/test/recursive-template-dependencies-test.ts
chriseppstein: Why is this `@amiller-gh` instead of `@css-blocks`?

packages/@css-blocks/ember-cli/tests/dummy/app/styles/node-modules-resolution.block.css
amiller-gh: Because the un-resolvable path throws a build time error, killing the test.

Actually, on second thought, this was included in here from when I pulled these fixes out of #196 into a separate PR. It shouldn't be in here. Will remove this route entirely for now.

packages/@css-blocks/ember-cli/tests/dummy/lib/in-repo-addon/package.json
amiller-gh: So, turns out npm and yarn don't consider beta releases matching when you specify `*`. It broke the build on fresh checkout because they pulled down a local version of 0.19.0. Once we get off the `0.20.0-beta` release line we can change these back to `*`.

amiller-gh: I'd like to release `0.20.0` in ~1w after we've integrated our first CSS Blocks component.

amiller-gh: And, for some reason I can't reply on your last comment below so I'll do it here.

`@amiller-gh/glimmer-analyzer` is a fork of `glimmer-analyzer` with some feature additions (dynamic config, package.json lookup path options, etc) @tomdale has expressed interest in `glimmer-analyzer` being pulled in to the `@glimmer` namespace, so I don't think our fork of it should live under `@css-blocks`, since it will hopefully disappear in the near future once we do the migration work. I didn't want to clutter up the `@css-blocks` namespace with packages we're planning to deprecate – I have far fewer concerns about the clutter in my personal github username's namespace 🙂

PR #201, opened 2018-10-02
Fixed bold "Future Feature: ..." notes

liamross:
There were spaces after the text and before the "**", causing it to render as

**Future Feature: ... **

instead of

**Future Feature: ...**


amiller-gh:
Thank you very much!!

PR #210, opened 2018-11-07
feat: Extended @block syntax. Issue #192.

amiller-gh:
- Rename @block-reference to @block
 - Default @block imports from a Block module.
 - Named @block imports from a Block module.
 - Default Block exports from a module.
 - Named Block exports from a module.
 - default as a reserved name.
 - Introduced simple parser for import/export parsing.
 - Split apart parser tests into logical packages.


amiller-gh:
@chriseppstein, still requires a few more tests to validate import/export error cases, but otherwise addresses the first half of functionality defined in https://github.com/linkedin/css-blocks/issues/192. Wanted to get it up here for early eyes on it. `node_modules` resolution will come in a separate PR 👍

amiller-gh:
Updated with more tests 👍 This is ready for review.

packages/@css-blocks/core/src/BlockParser/features/import-blocks.ts
```diff
@@ -0,0 +1,121 @@
...
+  // push to block references Promise array.
+  root.walkAtRules(BLOCK_IMPORT, (atRule: postcss.AtRule) => {
+    // <blocks-list> from <block-path>
+    let imports = atRule.params;
+
+    // <default-block> | <named-blocks> | <default-block> " , " <named-blocks> | <named-blocks> " , " <default-block>
+    let blockList = imports.split(FROM_EXPR)[0];
+
+    // ' " ' <any-value> ' " ' | " ' " <any-value> " ' ", then strip quotes.
+    let blockPath = stripQuotes(imports.split(FROM_EXPR)[1] || "");
```
chriseppstein: We shouldn't call split twice here. Should store the left and right side of the result at the first call then check against it.

amiller-gh: Mmm.

```js
let [blockList, blockPath] = imports.split(FROM_EXPR).map(stripQuotes);
```

amiller-gh: Updated

amiller-gh: Updated

chriseppstein:
@amiller-gh  Great work. I've merged this to master.

PR #213, opened 2018-11-29
[docs] Add repository/homepage to public package.jsons

mike-north:
Some of these fields were missing or out of date. This will make it easier to reach the source code and relevant readme from each package's respective [npm page](https://www.npmjs.com/package/@css-blocks/ember-cli)


chriseppstein:
@mike-north Thanks!

PR #218, opened 2018-12-06
chore: Cleanup shared test utils package.

amiller-gh:
- Move @css-blocks/build => @css-blocks/test-utils.
 - Move @css-blocks/test-utils to devDependencies in JSX integration.
 - Clean up exports API of test-utils so we don't need deep requires.
 - Clean up typings of mock function in test-utils so we don't need Object.apply.
 - Add README for test-utils.
 - Rename transient-fs to mock-fs.
 - Move _files and _directories private properties to truly private module scoped contants.


amiller-gh:
`@css-blocks/build` was a misnomer :) Also, made some small package structure and typing changes for cleaner APIs. Now that we have a shared test-utils package, I plan to also use it for the upcoming perf smoke testing infrastructure 👍

packages/@css-blocks/core/src/Analyzer/ElementAnalysis.ts
```diff
@@ -851,6 +851,7 @@ function mapChoiceClasses(
   configuration: ResolvedConfiguration,
   map: Map<string, Style>,
   includeAbsent: boolean,
+  /* tslint:disable-next-line */
```
amiller-gh: So... this has been throwing conflicting type errors in my IDE. Not sure how to fix. It either yells at me for expecting a newline and a comma after the param, or yells at me for adding a comma after a rest param. Any ideas on why?

packages/@css-blocks/jsx/package.json
```diff
@@ -40,6 +40,7 @@
     "access": "public"
   },
   "devDependencies": {
+    "@css-blocks/test-utils": "^0.20.0-beta.4",
```
amiller-gh: This wouldn't have published correctly otherwise :)

packages/@css-blocks/jsx/test/analyzer/class-states-objstr-test.ts
```diff
@@ -1,4 +1,5 @@
-import mock from "@css-blocks/build/dist/src/testing/transient-fs";
+import { mock } from "@css-blocks/test-utils";
```
amiller-gh: So much cleaner! 🎉

packages/@css-blocks/core/src/Analyzer/ElementAnalysis.ts
chriseppstein: yeah, I get that too. I think there's a bug in our custom linter due to a change in how rest parameters are modeled in the TS AST.

packages/@css-blocks/jsx/package.json
chriseppstein: good catch.

packages/@css-blocks/test-utils/package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "@css-blocks/build",
+  "name": "@css-blocks/test-utils",
   "private": true,
   "version": "0.20.0-beta.4",
-  "description": "Build-time support for @css-blocks/*.",
-  "main": "dist/src/index.js",
+  "description": "Shared test utilities for @css-blocks/*.",
```
chriseppstein: I had intended this to be a place for other types of build support, but we can cross that bridge when we get there.

packages/@css-blocks/test-utils/src/mock-fs/index.ts
```diff
@@ -6,14 +6,16 @@ export interface FileContent {
   [filename: string]: string | FileContent;
 }
 
-export interface TransientFiles {
+export interface FileMocker {
```
chriseppstein: a mock is a fake object that takes the place of a real object. This is a misnomer. I had kept the name "mock" at the call sites because I expected this to be a temporary work-around and that we would change back to mock-fs when it's working.

PR #219, opened 2018-12-06
feat: CSS Blocks Incremental Caching Build Strategy [WIP]

amiller-gh:
Major work in progress 🙂 By the end of this it should enable incremental builds. Will write a more detailed cache busting API description in this PR when I'm done exploring. For now, I'm mostly teasing apart tightly coupled dependencies and cleaning up tests to lay the groundwork, but want early eyes on some of these changes.


packages/@css-blocks/core/test/util/postcss-helper.ts
```diff
@@ -34,46 +34,25 @@ class Plugin {
...
   };
-  cssBlocks.OutputMode = OutputMode;
-  cssBlocks.CssBlockError = errors.CssBlockError;
-  cssBlocks.InvalidBlockSyntax = errors.InvalidBlockSyntax;
-  cssBlocks.MissingSourcePath = errors.MissingSourcePath;
-  return cssBlocks;
-}
-
-export = makeApi();
+};
```
chriseppstein: Honestly, this was when I thought CSS Blocks could (meaningfully) be a postcss plugin. This API should just be removed.

packages/@css-blocks/core/src/BlockFactory/BlockFactory.ts
```diff
@@ -0,0 +1,228 @@
...
+    this.preprocessQueue = new PromiseQueue(this.configuration.maxConcurrentCompiles, (item: PreprocessJob) => {
+      return item.preprocessor(item.filename, item.contents, this.configuration);
+    });
+  }
+
+  reset() {
+    this.blocks = {};
+    this.paths = {};
+    this.promises = {};
+    this.blockNames = {};
```
chriseppstein: Shouldn't this call `this.importer.purgeAll()`? If not, let's add a comment here so that it's documented why it doesn't do it for later readers who might also think it's a bug.

packages/@css-blocks/core/src/Importer/Importer.ts
```diff
@@ -0,0 +1,186 @@
...
+ * All importers enforce automatic memoization of `Importer.identifier()` and
+ * `Importer.import()` calls for improved performance and build consistency.
+ * Consumers may selectively purge these caches (ex: for incremental or watched
+ * builds) using the `Importer.purgeIdent()` and `Importer.purgeAll()` methods.
+ */
+export abstract class Importer {
+
+  private _identCache: Map<string, FileIdentifier> = new Map();
+  private _reverseIdentCache: Map<FileIdentifier, string> = new Map();
+  private _fileCache: Map<FileIdentifier, ImportedFile> = new Map();
```
chriseppstein: I think the file cache is the only thing we should attempt to cache generically. Caching idents is going to be too sparse because the path is scoped by the from identifier.

For large projects the file cache could grow unreasonably large without much benefit. In this case, a [LRU cache][node-lru-cache] can cap the memory usage to a configurable value and evict things from the cache that are either stale or which just don't appear to be accessed very often. We should use this to make sure we can set an upper bound to the amount of memory we devote to caching files.

[node-lru-cache]: https://github.com/isaacs/node-lru-cache

packages/@css-blocks/core/src/Importer/Importer.ts
```diff
@@ -0,0 +1,186 @@
...
+    this._fileCache.delete(ident);
+    return true;
+  }
+
+  /**
+   * Purge all cached idents from the `Importer` caches.
+   *
+   * @returns True when idents successfully purged.
+   */
+  purgeAll(): boolean {
```
chriseppstein: I don't think the name `purgeAll` is descriptive enough. I like `invalidateCache(identifier?: Identifier)`. Which would invalidate the entire cache if no identifier is provided or the cache value(s) associated with a specific identifier.

packages/@css-blocks/core/src/BlockTree/Block.ts
```diff
@@ -77,16 +76,28 @@ export class Block
...
+
+  /**
+   * Sets `timestamp` value of this `Block`. Block timestamps are set by `BlockFactory` after
+   * compilation to reflect the time at which its backing file contents have been read. Used
+   * for build invalidation.
+   * @prop  timestamp  number  The timestamp of this `Block`.
+   */
+  public setTimestamp(timestamp: number) {
+    if (!Number.isNaN(this._timestamp)) { throw new CssBlockError("Can not set Block timestamp more than once."); }
+    this._timestamp = timestamp;
```
chriseppstein: why isn't this done as a setter?

chriseppstein:
Where are blocks going to be invalidated because of dependencies between them?

packages/@css-blocks/core/src/Importer/NodeJsImporter.ts
```diff
@@ -124,13 +125,14 @@ export class NodeJsImporter implements Importer {
...
 
-  async import(identifier: FileIdentifier, config: ResolvedConfiguration): Promise<ImportedFile> {
+  async getImport(identifier: FileIdentifier, config: ResolvedConfiguration): Promise<ImportedFile> {
     let contents = await readFile(identifier, "utf-8");
     return {
       syntax: this.syntax(identifier, config),
       identifier,
       defaultName: this.defaultName(identifier, config),
       contents,
+      timestamp: Date.now(),
```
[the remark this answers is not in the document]
stefanpenner: It may be applicable or may not be, we have found the following [class](https://github.com/stefanpenner/fs-tree-diff/blob/master/lib/entry.js#L5) and [comparison function](https://github.com/stefanpenner/fs-tree-diff/blob/65a80920c2fb051f4cdab359afc9626257b81f6c/lib/index.js#L197-L212) to be handy abstractions for dealing with representing files on disk.

Also another example, where we have a  similar construct, but which is uses to abstract if we do or do not cache a files content
: https://github.com/broccolijs/broccoli-concat/blob/master/lib/strategies/simple.js#L13-L37

packages/@css-blocks/core/src/Importer/Importer.ts
chriseppstein: @stefanpenner the cache of block files won't vary by orders of magnitude like you'd see in broccoli-concat. I suspect that 99% of block files will vary between 1kb and 10kb. The rare exception would be if someone was using a lot of inline assets in a block file via a preprocessor. Even then, the file would be 50k on the large size.

The LRU would measure files by their size, but it evict purely on recent access. We could certainly decide that files of size greater than some percentage of the of LRU cap would not be cached to avoid the crowding out issue you mentioned.

Another interesting cache strategy would be to use the actual statistics from the previous run to inform evictions from the cache. Such data could also be used to recommend a minimum cache size to avoid evicting items that likely would have had a cache hit.

PR #220, opened 2018-12-06
Use typescript project references.

chriseppstein:
Project references allow typescript to efficiently build and watch dependencies
across TypeScript project boundaries.

There is a new tsconfig.json in packages that builds all the (non-ember)
typescript with a single compiler. This makes incremental builds work for all packages
based on what has changed.

Another great feature that TypeScript supports is also now enabled:
Type definition maps. This means that code navigation across packages will now
take you directly to the source instead of to the intermediate type
defintion from the compile. :tada:

The following new scripts are available:

* Top Level:
  * `yarn compile` - Build (or re-build) all typescript.
  * `yarn compile:clean` - Remove all compiled output and then compile.
  * `yarn compile:watch` - Start the typescript watcher for all
    typescript packages.
  * `yarn test:all` - Run all tests. This assumes the typescript has
    already been compiled.

* In each package:
  * `yarn compile` - Builds all typescript for the package and, if
    necessary, the local typescript projects that it depends on.
  * `yarn test` - Compile and run tests.
  * `yarn test:runner` - Run tests without compiling.


packages/@css-blocks/jsx/tsconfig.json
```diff
@@ -21,5 +22,10 @@
     "dist",
     "test/fixtures",
     "node_modules"
+  ],
+  "references": [
```
amiller-gh: Silly question because I don't know anything about the new typescript project refs feature, because these configs inherit from the base `tsconfig.json`, which includes all packages in its `references` section, are these package specific references needed?

chriseppstein: Those references are not in the `tsconfig.json` file that these inherit from.

In general, I think there's a tsconfig refactor that's needed. We might want to wait until 3.2 though: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#tsconfigjson-inheritance-via-nodejs-packages

PR #233, opened 2019-03-02
feat(glimmer): {{link-to}} integration.

amiller-gh:
## Problem 
In Ember, the built in `{{link-to}}` component internally applies an `active`, `loading` or `disabled` class when the application's current active route matches the route, is loading, or the link is disabled. In CSS Blocks, we need a method to style these internally managed `{{link-to}}` states. 

## Solution
The built in `{{link-to}}` helper has an attributes that enable developers to override the active, loading and disabled classes used by the component. A small change to the Glimmer analyzer and rewriter enable us to rewrite `{{link-to}}` like any other elements, but allow CSS Blocks to pass custom state styles.

Any `{{link-to}}` component discovered with one or more CSS Blocks classes applied to it will automatically have any `[state|active]`, `[state|loading]`, or `[state|disabled]` styles rewritten in to the component's `activeClass` hash value. Ex:

```css
/* stylesheet.css */
.link { color: red; }
.link[state|active] { color: green; }
.link[state|loading] { color: yellow; }
.link[state|disabled] { color: gray; }
```
```hbs
<div>
  {{link-to "My Link" "my-route" class="link"}}
</div>
```

is rewritten to

```css
/* stylesheet.css */
.link { color: red; }
.link--active { color: blue; }
.link--loading { color: yellow; }
.link--disabled { color: gray; }
```
```hbs
<div>
  {{link-to 
    "My Link" 
    "my-route" 
    class="link" 
    activeClass="link--active"
    loadingClass="link--loading"
    disabledClass="link--disabled"
  }}
</div>
```

The `style-if` and `style-unless` helpers are available to be used on `link-to` as well: 

```hbs
<div>
  {{link-to "My Link" "my-route" class=(style-if data "link")}}
</div>
```

Class inheritance and external block references work as expected.

## Implementation
The implementation internally generalizes the concept of `Ember Built-Ins`. In `@css-blocks/glimmer/src/BuiltIns.ts` we can define any number of components we want to treat like native elements. These Built-Ins may define zero to many state attribute mappings that will re-map the output classes of a Block to the component's defined style interface. We may want to consider using a similar implementation to enable user-defined component style interface mappings, or implement a Block passing feature (allow users to pass a custom implemented Block to a CSS Blocks component). For now though, the concept remains an internal construct.

### DIFF
--- a/packages/@css-blocks/ember-cli/tests/dummy/app/templates/ember-builtins.hbs
+++ b/packages/@css-blocks/ember-cli/tests/dummy/app/templates/ember-builtins.hbs
@@ -1,7 +1,4 @@
 <div>
   {{#link-to "index" class="link-to-helper" id="link-to-helper"}}Inactive Link To Helper{{/link-to}}
   {{#link-to "ember-builtins" class="link-to-helper" id="link-to-helper-active"}}Active Link To Helper{{/link-to}}
-
-  {{input type="text" class="input-helper" id="input-helper"}}
-  {{textarea type="text" class="textarea-helper" id="textarea-helper"}}
 </div>
--- a/packages/@css-blocks/ember-cli/tests/dummy/app/templates/ember-builtins.hbs
+++ b/packages/@css-blocks/ember-cli/tests/dummy/app/templates/ember-builtins.hbs
@@ -1,5 +1,7 @@
-{{#link-to "index" class="link-to-helper" id="link-to-helper"}}Link To Helper{{/link-to}}
-{{#link-to "ember-builtins" class="link-to-helper" id="link-to-helper-active"}}Active Link To Helper{{/link-to}}
+<div>
+  {{#link-to "index" class="link-to-helper" id="link-to-helper"}}Inactive Link To Helper{{/link-to}}
+  {{#link-to "ember-builtins" class="link-to-helper" id="link-to-helper-active"}}Active Link To Helper{{/link-to}}
 
-{{input type="text" class="input-helper" id="input-helper"}}
-{{textarea type="text" class="textarea-helper" id="textarea-helper"}}
+  {{input type="text" class="input-helper" id="input-helper"}}
+  {{textarea type="text" class="textarea-helper" id="textarea-helper"}}
+</div>
--- a/packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
+++ b/packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
@@ -72,9 +72,9 @@ export class ElementAnalyzer {
     return this.cssBlocksOpts.importer.debugIdentifier(this.block.identifier, this.cssBlocksOpts);
   }
 
-  private newElement(node: AnalyzableNodes, storeConditionals: boolean): TemplateElement {
+  private newElement(node: AnalyzableNodes, forRewrite: boolean): TemplateElement {
     let label = isElementNode(node) ? node.tag : node.path.original as string;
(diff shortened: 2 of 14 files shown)


packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -173,17 +229,18 @@ export class ElementAnalyzer {
               throw cssBlockError(`{{${helperType}}} expects a string literal as its third argument.`, elseBranch, this.template);
             }
           }
-          if (analysis.storeConditionals) {
-            analysis.element.addDynamicClasses(dynamicClasses(condition, whenTrue, whenFalse));
+          if (forRewrite) {
+            // element.addDynamicClasses(dynamicClasses(condition, whenTrue, whenFalse));
```
chriseppstein: why is this commented out?

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -173,17 +229,18 @@ export class ElementAnalyzer {
...
-            analysis.element.addDynamicClasses(dynamicClasses(null, whenTrue, whenFalse));
+            element.addDynamicClasses({ condition: null, whenTrue, whenFalse });
           }
 
         } else {
           throw cssBlockError(`Only {{style-if}} or {{style-unless}} helpers are allowed in class attributes.`, node, this.template);
         }
       } else {
-        assertNever(statement);
+        throw cssBlockError(`Only string literals, {{style-if}} or {{style-unless}} are allowed in class attributes.`, node, this.template);
```
chriseppstein: for this error and the one above `"helpers are allowed in class attributes"` seems strange to me. wouldn't `"helpers are allowed to use class attributes"` be more accurate?

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -1,26 +1,24 @@
...
 
 import { GlimmerAnalysis } from "./Analyzer";
+import { getEmberBuiltInStates, isEmberBuiltIn } from "./EmberBuiltins";
 import { ResolvedFile } from "./Template";
 import { cssBlockError } from "./utils";
 
-export type TernaryExpression = AST.Expression;
-export type StringExpression = AST.MustacheStatement | AST.ConcatStatement;
+export type TernaryExpression = AST.Expression | null;
+export type StringExpression = AST.MustacheStatement | AST.ConcatStatement | null;
```
chriseppstein: Please add code comments to these explaining the situation where `null` occurs.

packages/@css-blocks/ember-cli/tests/dummy/app/templates/ember-builtins.hbs
```diff
@@ -1,5 +1,7 @@
-{{#link-to "index" class="link-to-helper" id="link-to-helper"}}Link To Helper{{/link-to}}
-{{#link-to "ember-builtins" class="link-to-helper" id="link-to-helper-active"}}Active Link To Helper{{/link-to}}
+<div>
+  {{#link-to "index" class="link-to-helper" id="link-to-helper"}}Inactive Link To Helper{{/link-to}}
+  {{#link-to "ember-builtins" class="link-to-helper" id="link-to-helper-active"}}Active Link To Helper{{/link-to}}
```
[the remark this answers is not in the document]
amiller-gh: Yup, by virtue of the route matching – thats the magic that is `{{link-to}}`

PR #235, opened 2019-03-11
fix(webpack): Compatibility w/ webpack-dev-server.

### LINKED ISSUE #163: Analyzer not finding correct entry path with webpack-dev-server
Hi. I tried to install css-blocks in our environment that has Flow-typed React 16 code running on Webpack 3 and Babel 6. We're using webpack-dev-server for serving the app in dev. It seems the Analyzer plugin does not find the correct entry point but thinks the webpack-dev-server folder is the root folder (error below)
`
ERROR in [css-blocks] JSXParseError: Cannot read JSX entry point file /Users/myname/Documents/myapp/node_modules/webpack-dev-server/client/index.js?http:/localhost:3000: ENOENT: no such file or directory, open '/Users/myname/Documents/myapp/node_modules/webpack-dev-server/client/index.js?http:/localhost:3000' (/Users/myname/Documents/myapp/node_modules/webpack-dev-server/client/index.js?http:/localhost:3000)
`

Correct path would be:
 `/Users/myname/Documents/myapp/client/main.js`

I have this config in webpack config:
```js
  const analyzerInstance = new CssBlocks.Analyzer('client/main.js', {
    baseDir: '/Users/myname/Documents/myapp',
    types: 'flow',
  });
```
but it doesn't seem to have effect.

ps. Both our app and webpack-dev-server happen to use `client` folder as part of the path.  So `client/index.js` is not mistyped path to our entry but the actual entry point of the dev server. (I was confused by this myself for a second)

nickiaconis:
This fix filters out fake paths injected by the dev server. Doing so prevents the throwing of ENOENT errors when the Analyzer tries to open those fake paths. Fixes #163.

I've tested this manually by linking it into a project that uses webpack-dev-server and observing that compilation succeeds where it had previously failed.

The test I added fails with `Module build failed: Error: No rewriter` as do all but one of the existing Plugin tests when unskipped. (The outlier fails with a different error.) Since the existing tests have been skipped, I figure it's okay to do the same with this one.

### DIFF
--- a/packages/@css-blocks/webpack/src/Plugin.ts
+++ b/packages/@css-blocks/webpack/src/Plugin.ts
@@ -117,6 +117,9 @@ export class CssBlocksPlugin
       entries = flatten<string>(objectValues(webpackEntry));
     }
 
+    // Zomg webpack-dev-server, injecting fake paths into the entry points.
+    entries = entries.filter(entry => !/\/webpack-dev-server\/|^webpack\/hot\/dev-server$/.test(entry));
+
     let pending: PendingResult = this.analyzer.analyze("", entries)
       // If analysis fails, drain our BlockFactory, add error to compilation error list and propagate.
       .catch((err: Error) => {
--- a/packages/@css-blocks/webpack/test/plugin-test.ts
+++ b/packages/@css-blocks/webpack/test/plugin-test.ts
@@ -42,6 +42,10 @@ export class PluginTest {
     });
   }
 
+  @skip @test "works with HotModuleReplacementPlugin"() {
+    return execTest("hello", undefined, "dev-server");
+  }
+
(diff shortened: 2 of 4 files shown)


packages/@css-blocks/webpack/src/Plugin.ts
```diff
@@ -117,6 +117,9 @@ export class CssBlocksPlugin
       entries = flatten(objectValues(webpackEntry));
     }
 
+    // Zomg webpack-dev-server, injecting fake paths into the entry points.
+    entries = entries.filter(entry => !/\/webpack-dev-server\/|^webpack\/hot\/dev-server$/.test(entry));
```
nickiaconis: Perhaps the regex needs to account for Windows-esque backslashes too?

PR #238, opened 2019-03-30
fix: Properly output conflict resolutions for shorthands.

amiller-gh:
- ConflictResolver now takes shorthands into consideration.
 - PropertyParser has special cases for border and font-family because of bugs in source package.
 - fix: Export syntax now allows for immediate re-export from file path.
 - Added more debug statements.
 - Fix Broccoli bug to only remove files from tmp directory output.

### DIFF
--- a/packages/@css-blocks/core/src/BlockTree/RulesetContainer.ts
+++ b/packages/@css-blocks/core/src/BlockTree/RulesetContainer.ts
@@ -7,12 +7,12 @@
  * `::after`). Every ruleset may provide resolution guidance against another `Style` object.
  */
 import { MultiMap, TwoKeyMultiMap } from "@opticss/util";
-import * as propParser from "css-property-parser";
 import { ParsedSelector, postcss } from "opticss";
 
 import { BLOCK_PROP_NAMES, SELF_SELECTOR, getResolution } from "../BlockSyntax";
 import { InvalidBlockSyntax } from "../errors";
 import { sourceLocation } from "../SourceLocation";
+import { expandProp } from "../util/propertyParser";
 
 import { Styles, isStyle } from "./Styles";
 export { Styles, BlockClass, AttrValue } from "./Styles";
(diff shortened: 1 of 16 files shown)


packages/@css-blocks/core/src/BlockTree/RulesetContainer.ts
```diff
@@ -32,25 +32,6 @@ export type Resolution<S extends Styles = Styles> = {
   resolution: S;
 };
 
-/**
```
amiller-gh: Moved to utils

packages/@css-blocks/core/src/importing/NodeJsImporter.ts
```diff
@@ -68,7 +68,12 @@ export class NodeJsImporter implements Importer {
 
     // If no alias found, test for a node_module resolution as a file path.
     try {
-      return require.resolve(importPath, { paths: [config.rootDir] });
+      const file = require.resolve(importPath, { paths: [config.rootDir] });
+      const extname = path.extname(file).slice(1);
+      const fileExts = { css: true, ...config.preprocessors };
+      if (!fileExts[extname]) { throw Error("Invalid Block File Extension"); }
```
amiller-gh: If we have resolved to a file – it might be a JS file! Ensure this is a CSS Blocks parsable file first.

packages/@css-blocks/core/src/util/propertyParser.ts
```diff
@@ -0,0 +1,66 @@
+import * as propParser from "css-property-parser";
```
amiller-gh: There are a lot of bugs in PropertyParser... rather than open a new PR against a largely auto-generated project, I've opted to special case them here for now. I've opened issues against css-property-parser in the mean time.

packages/@css-blocks/core/test/block-inheritance-test.ts
```diff
@@ -195,7 +195,9 @@ export class BlockInheritance extends BEMProcessor {
         result.css.toString(),
         ".sub__nav { margin: 15px; }\n" +
         ".base__nav.sub__nav { margin: 15px; }\n" +
+        ".base__nav + .base__nav.sub__nav { margin: 15px; }\n" +
```
amiller-gh: Found bugs in resolution! Well... kind of. It was doing the right thing for the wrong reasons. This can be handled by opticss, but theres no need to output full selector enumeration with the current CSS Blocks rules. Kind of a bug, kind of not.

PR #245, opened 2019-04-26
Command Line Interface

chriseppstein:
The CLI is useful for testing css-blocks or doing basic compilation to BEM classes.


packages/@css-blocks/cli/src/index.ts
```diff
@@ -0,0 +1,110 @@
+import { BlockFactory, CssBlockError, Preprocessors } from "@css-blocks/core";
```
amiller-gh: Add comments to this file then looks good to me.

packages/@css-blocks/cli/src/index.ts
```diff
@@ -0,0 +1,110 @@
...
+import chalk = require("chalk");
+import fs = require("fs");
```
amiller-gh: Doesn't `fs-extra` have a promisified version of writeFile you can use?

packages/@css-blocks/cli/src/index.ts
chriseppstein: ✅ Comments added.

packages/@css-blocks/cli/src/index.ts
chriseppstein: yep! Using that now.

PR #247, opened 2019-04-29
fix: Over-zealous conflicts from inherited in-stylesheet compositions.

amiller-gh:


packages/@css-blocks/core/src/Analyzer/validations/property-conflict-validator.ts
```diff
@@ -111,16 +125,18 @@ function recursivelyPruneConflicts(prop: string, conflicts: ConflictMap): Rulese
...
+    let decl = rule.declarations.get(prop);
+    let nodes: postcss.Rule[] | postcss.Declaration[] =  decl ? decl.map((d) => d.node) : [rule.node];
+    for (let node of nodes) {
+      let line = node.source.start && `:${node.source.start.line}`;
+      let column = node.source.start && `:${node.source.start.column}`;
+      out.add(`    ${rule.style.asSource(true)} (${rule.file}${line}${column})`);
+    }
   }
-  return out.join("\n");
+  return [...out].join("\n");
```
amiller-gh: If multiple conflicts with the same line in a set, only prints unique lines once. Reduces console noise.

packages/@css-blocks/core/src/BlockTree/BlockClass.ts
```diff
@@ -268,6 +268,21 @@ export class BlockClass extends Style<BlockClass, Block, Block, Attribute> {
     this._composedStyles.add({ style, conditions });
   }
 
+  /**
```
amiller-gh: Hm, this comment is wrong. Updating.

packages/@css-blocks/core/src/Analyzer/validations/property-conflict-validator.ts
```diff
@@ -111,16 +125,18 @@ function recursivelyPruneConflicts(prop: string, conflicts: ConflictMap): Rulese
...
-    out.push(`    ${rule.style.asSource(true)} (${rule.file}${line}${column})`);
+function printRulesetConflict(prop: string, rules: Ruleset<Style>[]) {
+  const out = new Set();
+  for (let rule of rules) {
+    let decl = rule.declarations.get(prop);
+    let nodes: postcss.Rule[] | postcss.Declaration[] =  decl ? decl.map((d) => d.node) : [rule.node];
+    for (let node of nodes) {
+      let line = node.source.start && `:${node.source.start.line}`;
+      let column = node.source.start && `:${node.source.start.column}`;
+      out.add(`    ${rule.style.asSource(true)} (${rule.file}${line}${column})`);
```
chriseppstein: doesn't this print `undefined` when `node.source.start` is undefined?

packages/@css-blocks/core/src/importing/NodeJsImporter.ts
```diff
@@ -51,20 +51,22 @@ export class NodeJsImporter implements Importer {
...
     from = from ? this.filesystemPath(from, config) : from;
     let fromDir = from ? path.dirname(from) : config.rootDir;
     let resolvedPath = path.resolve(fromDir, importPath);
     if (existsSync(resolvedPath)) { return resolvedPath; }
     debug(`No relative or absolute Block file discovered for ${importPath}.`);
 
-    // If not a real file, attempt to resolve to an aliased path instead.
-    let alias = this.aliases.find(a => importPath.startsWith(a.alias + path.sep));
+    // If not a real file, attempt to resolve to an aliased path instead, if present.
+    let alias = this.aliases.find(a => importPath.startsWith(a.alias) || importPath.startsWith(a.alias + path.sep));
```
chriseppstein: If `importPath` doesn't start with `a.alias` how can it start with `a.alias + path.sep`? The second part of this expression will always be false if it runs.

chriseppstein: I removed the redundant branch.

packages/@css-blocks/core/src/Analyzer/validations/property-conflict-validator.ts
chriseppstein: I set this to an empty string in that case.

packages/@css-blocks/core/src/BlockTree/BlockClass.ts
chriseppstein: I'm going to land this PR. Please update the comment on master.

PR #302, opened 2019-07-26
Update main.css

### LINKED ISSUE #222: TSDoc API display issue
### The following screenshot shows the visual bug and is found [here](https://css-blocks.com/api/modules/_css_blocks_jsx.html#babel_plugins):

screen shot 2018-12-10 at 1 20 15 pm

### The following lines are responsible for the style of this section:

https://github.com/linkedin/css-blocks/blob/c447f2f99978720f3e1b0143983207161557a571/packages/%40css-blocks/website/typedoc-theme/assets/css/main.css#L782

### Adding the following code would fix the issue:

```css
.tsd-panel > .tsd-signature {
  word-break: break-all;
}
```

It probably can't be added to the theme css which was linked above, but it could go into any overriding stylesheet (I took a look around but couldn't find one quickly or else I would have done it myself).

Additionally, this could be a bug involving how this data is being parsed as it could be solved by simply inserting spaces between each element of the array, which would no longer require the word-break style.

### Other notes:

Here's a link to the constant that it is documenting:

https://github.com/linkedin/css-blocks/blob/38eda83fa6ef7e63723ff4a75fbed50227d74d11/packages/%40css-blocks/jsx/src/options.ts#L11-L33

liamross:
Fixes #222

### The following screenshot shows the visual bug and is found [here](https://css-blocks.com/api/modules/_css_blocks_jsx.html#babel_plugins):

screen shot 2018-12-10 at 1 20 15 pm

### After fix

Screen Shot 2019-07-26 at 3 31 08 PM


chriseppstein:
@liamross Thank you for the fix!

Issue #248, opened 2019-04-29
Error Location should have a stack.


chriseppstein:
When an error occurs in parsing a dependent block, the error location needs to include the code locations of all the `@block` commands that got to that block from the entry source file.

PR #319, opened 2019-09-06
Block import stack

chriseppstein:
When a block doesn't compile due to a downstream error in a block reference, the locations of the `@block` at-rules that brought in the failing block file are now part of the error's public interface and are displayed in the CLI error output.

The error locations are source-mapped where appropriate.

import stack


chriseppstein:
This PR addresses #248.

PR #324, opened 2019-09-19
Language server spike

chriseppstein:
Most of this work was done by @jackson-dean. I adapted the example structure that vscode tooling creates so that it fits into our monorepo structure. More work is needed, but the basic development workflow works now.


chriseppstein:
Merged in as 4c156d1d6edc80945e92a1aa9434c25d831026dc.

PR #330, opened 2019-09-25
feat: Respect explicit exports for a block interface.

ramitha:
Previously, the block's interface constituted of all of its imported blocks as well as the classes and states defined on the block.
Now, the block's public interface will consist of only that which is explicitly exported using @export


chriseppstein:
Look good. Thanks!

Issue #332, opened 2019-09-26
RFC: Per-Block Namespaces


chriseppstein:
# Summary

Instead of using a single namespace `state` for the attributes, we will formalize the concept that the block classes and states live in a shared namespace that is defined by the block. While the namespace for a block will be the url of that block (`file://...`) the local references to block styles will use the locally assigned identifier to refer to it via a namespace instead of using a non-standard dotted notation.

This provides significant design benefits, improves the overall readability of the code, and reduces the amount of boilerplate syntax the authors must provide.

### Syntax Example

```css
/* section.block.css -- automatically associated with above template */
@export btn from "./buttons.block.css"; /* automatically defines the namespace associated with buttons.block.css and assigns a identifier of "btn" to it. */
:scope { /* ... */ }
:scope[enabled] { /* ... */ }
.foot { /* ... */ }
```

```css
/* button.block.css */
:scope { /* ... */ }
:scope:hover { /* ... */ }
:scope:active { /* ... */ }
:scope:disabled { /* ... */ }
.icon { /* ... */ }
.icon[inverse] { /* ... */ }
```

```hbs
{{!-- section.hbs--}}
<section block:scope block:enabled={{isEnabled}}>
  <button btn:scope>
    <div btn:class="icon" btn:inverse={{isInverse}}></div>
    {{value}}
  </button>
  <footer block:class="foot">
    Copyright (c) 2019 LinkedIn Corp.<br>
    All rights reserved.
  </footer>
</section>
```

# Motivation

Consistent feedback that I've gotten about CSS Blocks is that the state namespace is "weird" and "confusing".

We've used this namespace primarily as a signal that the attribute(s) belong to CSS Blocks instead of HTML so that we can confidently rewrite them.

While the word `state` was meant to evoke the visual state of the element, it can also be confusing with other "states" going on in a component. By using the name of the block, the code is more clearly associated with that block and so it reads better.

By taking over the html class attribute we've caused ourselves a lot of challenges in terms of legacy support that makes the gradual adoption of CSS Blocks more of a challenge than it needs to be. This change will allow legacy classes to coexist with css-blocks' classes without ambiguity so we do not lose our ability to clearly error when a block class isn't found.

# How do we teach this?

Fortunately there's not very many users of CSS Blocks yet 😅.

[...]

PR #336, opened 2019-09-29
Namespace per block.

chriseppstein:
RFC: https://github.com/linkedin/css-blocks/issues/332

### CSS Changes

- [x] The class selector will now be defined as selecting an attribute named `class` in the namespace of the block. (conceptual change, not requiring any stylesheet parsing changes at this time).
- [x] The default namespace for the attribute selector will be dom attributes in the namespace of the block. Therefore the `state|` namespace in the state attribute selector is no longer needed.
- [x] Global states are accessed via the attribute namespace instead of a fake tag name.
- [ ] The `:scope` selector will select an element with the attribute of `scope` that is in the namespace of the block.
- [x] States named "scope" will no longer be allowed.
- [x] non-block namespaces will be handled by registering common names of `html`, `svg`, and `math` and making them forbidden names for blocks.
- [ ] Other non-block namespaces can be registered through configuration.

### Handlebars Template Syntax Changes

- [x] The namespace identifier for the default block associated with the template is `block`.
- [ ] If no element has an attribute of `block:scope` then it is automatically assigned to the template's root element.
- [x] Instead of `class="foo other-block.bar"` we would now write `block:class="foo" other-block:class="bar"`.
- [x] The `class` attribute in html namespace is forbidden. (Later we can add configuration to allow it.)

### Documentation Changes
- [ ] Document namespace concept in tutorials/readmes.
- [x] Website syntax updates.


amiller-gh:
One more item to add to the TODO list: Make the word `block` a reserved word (if it's not already... I don't think it is).

Because you reference the "default" exported block as `block`, the compiler should throw if you try to `@export (whatever as block ) from "wherever.block.css"`.

amiller-gh:
Alt: You can just change the keyword `default` to `block` for the import/export parser. 

This may be confusing for those coming from JavaScript world though – would not recommend.

packages/@css-blocks/core/src/BlockParser/block-intermediates.ts
```diff
@@ -122,13 +114,16 @@ export function isRootNode(node: unknown): node is selectorParser.Pseudo {
 
 export const isClassNode = selectorParser.isClassName;
 
+export const RESERVED_NAMESPACES = new Set<string | undefined | true>(["html", "math", "svg"]);
+Object.freeze(RESERVED_NAMESPACES);
```
jackson-dean: is this left overs? it seems the freeze is called but not assigned to anything. It also doesn't seem that freeze prevents adding to a Set.

packages/@css-blocks/core/src/BlockSyntax/BlockSyntax.ts
```diff
@@ -70,3 +67,23 @@ export const ROOT_CLASS = ":scope";
...
 // The string `default` is used throughout the system to represent the default export of a block file.
 export const DEFAULT_EXPORT = "default";
+
+// The namespace used for referencing a default block from a template
+export const DEFAULT_NAMESPACE = "block";
+
+/**
+ * Names that a block cannot have as a namespace.
+ */
+export const RESERVED_NAMESPACES = new Set<string>([DEFAULT_EXPORT, "html", "svg", "math", "mathml", "xlink", "xml", "xmlns", "xhtml"]);
```
jackson-dean: I could just be missing something, but this looks suspiciously similar to the RESERVED_NAMESPACES const from the previous file. Are they the same or somehow different?

packages/@css-blocks/vscode/fixtures/app/styles/nav.block.css
```diff
@@ -1,3 +1,5 @@
+@export subnav from "./subnav.block.css";
```
jackson-dean: No issue, but I'm guessing these are just setup for future test? It doesn't appear these fixtures are used atm.

chriseppstein: I was using them for manual testing. Adding test cases is a good idea...

packages/@css-blocks/core/src/BlockSyntax/BlockSyntax.ts
chriseppstein: Good catch. I've standardized on this one.

packages/@css-blocks/core/src/BlockParser/block-intermediates.ts
chriseppstein: I removed this line.

PR #342, opened 2019-10-08
Public classes

ramitha:
Adding support for block-aliases for styles

### DIFF
--- a/packages/@css-blocks/core/src/Analyzer/Analysis.ts
+++ b/packages/@css-blocks/core/src/Analyzer/Analysis.ts
@@ -338,7 +338,7 @@ export class Analysis<K extends keyof TemplateTypes> {
    * @param options The plugin options that are used to parse the blocks.
    * @param postcssImpl The instance of postcss that should be used to parse the block's css.
    */
-  async deserialize (
+  static async deserialize (
     serializedAnalysis: SerializedAnalysis<keyof TemplateTypes>,
     blockFactory: BlockFactory,
     parent: Analyzer<keyof TemplateTypes>,
@@ -377,7 +377,7 @@ export class Analysis<K extends keyof TemplateTypes> {
     let elementNames = Object.keys(serializedAnalysis.elements);
     elementNames.forEach((elID) => {
       let data = serializedAnalysis.elements[elID];
-      let element = new ElementAnalysis<null, null, null>(data.sourceLocation || {start: POSITION_UNKNOWN}, this.reservedClassNames(), undefined, elID);
+      let element = new ElementAnalysis<null, null, null>(data.sourceLocation || {start: POSITION_UNKNOWN}, parent.reservedClassNames(), undefined, elID);
       analysis.elements.set(elID, element);
     });
 
--- a/packages/@css-blocks/core/src/Analyzer/Analysis.ts
+++ b/packages/@css-blocks/core/src/Analyzer/Analysis.ts
@@ -180,7 +180,7 @@ export class Analysis<K extends keyof TemplateTypes> {
     if (this.currentElement) {
       throw new Error("Internal Error: failure to call endElement before previous call to startElement.");
     }
-    this.currentElement = new ElementAnalysis<BooleanExpression, StringExpression, TernaryExpression>(this.reservedClassNames(), location, tagName, this.idGenerator.nextIdent());
+    this.currentElement = new ElementAnalysis<BooleanExpression, StringExpression, TernaryExpression>(location, this.reservedClassNames(), tagName, this.idGenerator.nextIdent());
     return this.currentElement;
   }
 
@@ -377,7 +377,7 @@ export class Analysis<K extends keyof TemplateTypes> {
     let elementNames = Object.keys(serializedAnalysis.elements);
     elementNames.forEach((elID) => {
       let data = serializedAnalysis.elements[elID];
-      let element = new ElementAnalysis<null, null, null>(this.reservedClassNames(), data.sourceLocation || {start: POSITION_UNKNOWN}, undefined, elID);
+      let element = new ElementAnalysis<null, null, null>(data.sourceLocation || {start: POSITION_UNKNOWN}, this.reservedClassNames(), undefined, elID);
       analysis.elements.set(elID, element);
     });
 
--- a/packages/@css-blocks/core/src/Analyzer/Analysis.ts
+++ b/packages/@css-blocks/core/src/Analyzer/Analysis.ts
@@ -180,7 +180,7 @@ export class Analysis<K extends keyof TemplateTypes> {
     if (this.currentElement) {
       throw new Error("Internal Error: failure to call endElement before previous call to startElement.");
     }
-    this.currentElement = new ElementAnalysis<BooleanExpression, StringExpression, TernaryExpression>(location, tagName, this.idGenerator.nextIdent());
+    this.currentElement = new ElementAnalysis<BooleanExpression, StringExpression, TernaryExpression>(this.reservedClassNames(), location, tagName, this.idGenerator.nextIdent());
     return this.currentElement;
   }
 
@@ -238,6 +238,19 @@ export class Analysis<K extends keyof TemplateTypes> {
    */
   referencedBlocks(): Block[] {
     return objectValues(this.blocks);
+  }
+
+  /**
+   * For now, returns all aliases referenced by this block and all the blocks they
+   * reference recursively. We can add more to this set in future
+   */
+  reservedClassNames(): Set<string> {
+    let aliases = new Set<string>();
+    let blocks = this.transitiveBlockDependencies();
+    blocks.forEach(block => {
+      block.getAllStyleAliases().forEach(alias => aliases.add(alias));
+    });
+    return aliases;
   }
 
   /**
@@ -325,7 +338,7 @@ export class Analysis<K extends keyof TemplateTypes> {
    * @param options The plugin options that are used to parse the blocks.
    * @param postcssImpl The instance of postcss that should be used to parse the block's css.
    */
-  static async deserialize (
+  async deserialize (
     serializedAnalysis: SerializedAnalysis<keyof TemplateTypes>,
     blockFactory: BlockFactory,
     parent: Analyzer<keyof TemplateTypes>,
@@ -364,7 +377,7 @@ export class Analysis<K extends keyof TemplateTypes> {
     let elementNames = Object.keys(serializedAnalysis.elements);
     elementNames.forEach((elID) => {
       let data = serializedAnalysis.elements[elID];
-      let element = new ElementAnalysis<null, null, null>(data.sourceLocation || {start: POSITION_UNKNOWN}, undefined, elID);
+      let element = new ElementAnalysis<null, null, null>(this.reservedClassNames(), data.sourceLocation || {start: POSITION_UNKNOWN}, undefined, elID);
       analysis.elements.set(elID, element);
     });
 
--- a/packages/@css-blocks/core/src/BlockCompiler/index.ts
+++ b/packages/@css-blocks/core/src/BlockCompiler/index.ts
@@ -3,6 +3,7 @@ import { postcss } from "opticss";
 
 import { Analyzer } from "../Analyzer";
 import {
+  BLOCK_ALIAS,
   BLOCK_AT_RULES,
   BLOCK_DEBUG,
   BLOCK_PROP_NAMES_RE,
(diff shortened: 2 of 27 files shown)


packages/@css-blocks/core/src/BlockParser/features/construct-block.ts
```diff
@@ -105,6 +107,14 @@ export async function constructBlock(configuration: Configuration, root: postcss
   return block;
 }
 
+function addStyleRules(style: AttrValue | BlockClass, rule: postcss.Rule, tuple: Set<[Style, postcss.Rule]>): void {
+  rule.walkDecls(BLOCK_ALIAS, decl => {
+    style.setStyleAliases(new Set(decl.value.split(/\s+/).map(stripQuotes)));
+    decl.remove();
```
chriseppstein: Mutations to the stylesheet should occur in the `BlockCompiler`, not the parser.

packages/@css-blocks/core/src/BlockTree/Block.ts
```diff
@@ -393,6 +393,22 @@ export class Block
...
+   * @returns Array of style aliases.
+   */
+  getAllStyleAliases(shallow?: boolean): Set<string> {
+    let result = new Set<string>();
+    for (let blockClass of this.classes) {
+      result = new Set([...result, ...blockClass.getStyleAliases()]);
+    }
+    if (!shallow && this.base) {
+      result = new Set([...result, ...this.base.getAllStyleAliases(shallow)]);
+    }
```
chriseppstein: This does a lot of allocations to save a few extra lines of code. I think this should loop more and call `result.add()` on the individual strings.

Also, there's no need to pass `shallow` along to the recursive `getAllStyleAliases`, it's definitely `false` in that situation.

packages/@css-blocks/core/src/BlockTree/Style.ts
```diff
@@ -61,11 +64,47 @@ export abstract class Style<
...
   }
 
+  /**
+   * Adds a set of aliases to the to this object
+   */
+  public setStyleAliases(aliases: Set<string>): void {
+    if (aliases.size) {
+      // clear the existing styleAliases if they exist
+      if (this._styleAliases) {
+        this._styleAliases.clear();
```
chriseppstein: In this branch of the code it seems like we don't set the aliases to the value that is passed in. I don't understand why we need to branch on the existing style aliases presence if we're just setting the field to the value passed in... it seems like the same assignment works for both cases.

packages/@css-blocks/core/test/Block/style-alias-test.ts
```diff
@@ -0,0 +1,105 @@
...
+    return this.process(filename, inputCSS, config).then(() => assert(false, "noError was thown")).catch(err => {
+      assert.equal(err.name, "CssSyntaxError");
+    });
+  }
+
+  @test "should parse quoted aliases containing special characters"() {
+    let { config } = setupImporting();
+
+    let filename = "foo/bar/a-block.css";
+    let inputCSS = `:scope { block-alias: a-block-alias1 "a-block-alias2-;-with"; color: red; }
```
chriseppstein: I think we should only allow legal classnames as aliases so this should be an error. Note: there's a Regexp somewhere in the project for validating a classname...

packages/@css-blocks/core/test/opticss-test.ts
```diff
@@ -56,8 +56,8 @@ export class TemplateAnalysisTests {
     let self = this;
     let filename = "blocks/foo.block.css";
     let css = clean`
-      :scope { color: blue; font-size: 20px; }
-      :scope[state|foo] { color: red; }
+      :scope { block-alias: my-foo-alias "somethingwith,splcharacters;"; color: blue; font-size: 20px; }
```
chriseppstein: same comment as above.

packages/@css-blocks/core/test/opticss-test.ts
```diff
@@ -88,15 +88,26 @@ export class TemplateAnalysisTests {
...
+          id: false,
+          class: true,
+          omitIdents: {
+            id: [],
+            class: [],
+          },
+        },
+        analyzedAttributes: [],
+        analyzedTagnames: false,
+      });
```
chriseppstein: This change seems to not have any real difference to it.

packages/@css-blocks/glimmer/test/fixtures/styled-app/src/ui/components/with-block-aliases/stylesheet.css
```diff
@@ -0,0 +1,16 @@
...
+
+.world {
+    block-alias: stylesheet__world--thick;
+    border: 1px solid black;
+}
+
+.world[state|thick] {
+    block-alias: my-alias-for-state;
+    border-width: 3px;
+}
```
chriseppstein: I hadn't considered this use case where a style is aliased to the default class for a different style :| Can you please explain how the code handles this?

ramitha: I had to take care of this explicitly inside of the `BlockClass` while generating `cssClass` for the style node. If the generated BEM name is a style alias on the block, we append a guid to the generated name.

chriseppstein: ok. I think I understand the intent here... by keeping the styles associated with a private class name, the alias class can't be used to style elements that aren't rewritten by CSS Blocks.

I'm on board with this general goal, but it's tricky. The unique class guarantee is currently only implemented for block classes, but not for block state attributes. Furthermore, this only considers the current block, but the alias might also be a name used by a different block in the same build, and because CSS Blocks aims to make whole-application guarantees about naming, if we implement this, I think it should take all blocks in the build into account.

To this end, I think we should change `cssClass` to accept a second parameter that is a collection of reserved classnames that the method cannot use. Then we can collect all the style aliases of all the blocks after they've been parsed, and pass them to the block compiler to ensure that all the class names we generate for the entire build do not conflict with _any_ public alias.

```ts
public cssClass(config: ResolvedConfiguration, reservedClassNames: Set<string>): string;
```

packages/@css-blocks/core/src/BlockParser/features/construct-block.ts
```diff
@@ -107,10 +107,20 @@ export async function constructBlock(configuration: Configuration, root: postcss
...
+function addStyleRules(configuration: Configuration, block: Block, rule: postcss.Rule, file: string, style: AttrValue | BlockClass, tuple: Set<[Style, postcss.Rule]>): void {
   rule.walkDecls(BLOCK_ALIAS, decl => {
-    style.setStyleAliases(new Set(decl.value.split(/\s+/).map(stripQuotes)));
-    decl.remove();
+    let aliases: Set<string> = new Set();
+    decl.value.split(/\s+/).map(alias => {
+      let cleanedAlias = stripQuotes(alias);
+      if (!CLASS_NAME_IDENT.test(cleanedAlias)) {
+        throw new errors.InvalidBlockSyntax(
+          `Illegal block-alias in export. "${alias}" is not a legal CSS identifier.`,
```
chriseppstein:
```suggestion
          `The block-alias "${alias}" must be a legal CSS class name.`,
```

packages/@css-blocks/core/src/BlockTree/Block.ts
```diff
@@ -395,16 +395,12 @@ export class Block
...
    */
-  getAllStyleAliases(shallow?: boolean): Set<string> {
+  getAllStyleAliases(): Set<string> {
     let result = new Set<string>();
     for (let blockClass of this.classes) {
-      result = new Set([...result, ...blockClass.getStyleAliases()]);
-    }
-    if (!shallow && this.base) {
-      result = new Set([...result, ...this.base.getAllStyleAliases(shallow)]);
+      blockClass.getStyleAliases().forEach(alias => result.add(alias));
```
chriseppstein: This method doesn't include the style aliases on state attribute values.

packages/@css-blocks/glimmer/test/fixtures/styled-app/src/ui/components/with-block-aliases/stylesheet.css
```diff
@@ -0,0 +1,16 @@
...
+    color: red;
+}
+
+.world {
+    block-alias: stylesheet__world--thick;
+    border: 1px solid black;
+}
+
+.world[state|thick] {
+    block-alias: my-alias-for-state;
```
chriseppstein: This one should probably conflict with the class for `:scope`.

packages/@css-blocks/glimmer/test/fixtures/styled-app/src/ui/components/with-block-aliases/stylesheet.css
ramitha: I see what you mean. I'll make changes to account for this.

packages/@css-blocks/core/src/BlockTree/Block.ts
```diff
@@ -454,28 +469,28 @@ export class Block
...
+  rewriteSelectorNodes(nodes: selectorParser.Node[], config: ResolvedConfiguration, reservedClassNames: Set<string>): selectorParser.Node[] {
     let newNodes: selectorParser.Node[] = [];
     for (let i = 0; i < nodes.length; i++) {
       let node = nodes[i];
       let result = this.nodeAsStyle(node);
       if (result === null) {
         newNodes.push(node);
       } else {
-        newNodes.push(selectorParser.className({ value: result[0].cssClass(config) }));
+        newNodes.push(selectorParser.className({ value: result[0].cssClass(config, reservedClassNames)}));
```
chriseppstein: Looking at this code, I'm concerned by the fact that the value for `cssClass` will end up different upon different invocations depending on what value is passed for `config` and `reservedClassNames`. Once parse and analysis is complete and compilation/rewrite is about to occur, the CSS Class for each style must be immutable. But this code structure doesn't express that fact (I know that this change didn't introduce this problem... it just exacerbated it).

I think a Block should have a method that puts it into a state of having determined all the classnames to be used. E.g. `block.determineClassNames(config, reservedClassNames)` which would in turn precompute all the class names for all the styles in the block and then `style.cssClass()` can just be a simple accessor (it would raise an error if the classname wasn't yet determined).

I think this will also have the effect of cleaning up _a lot_ of code that is passing around `config` and `reservedClassNames`.

packages/@css-blocks/core/src/BlockTree/Style.ts
```diff
@@ -58,14 +61,43 @@ export abstract class Style<
...
+   */
+  public setStyleAliases(aliases: Set<string>): void {
+    this._styleAliases = aliases;
+  }
+
+  /**
+   * Returns the alisses on this object
+   */
+  public getStyleAliases(): Set<string> {
+    return this._styleAliases || new Set();
```
chriseppstein: Most things don't have a style alias so this makes a lot of objects unnecessarily. let's factor out a singleton constant like `const NO_STYLE_ALIASES = new Set<string>()` and return that same instance any time there's no style alias.

ramitha: Fixed.

packages/@css-blocks/core/src/BlockTree/Block.ts
ramitha: This is a great point. I'm going to address it as part of another PR.

packages/@css-blocks/core/src/BlockCompiler/index.ts
```diff
@@ -54,14 +54,17 @@ export class BlockCompiler {
...
     // Resolve inheritance based conflicts
     resolver.resolveInheritance(root, block);
     root.walkRules((rule) => {
       let parsedSelectors = block.getParsedSelectors(rule);
-      rule.selector = parsedSelectors.map(s => block.rewriteSelectorToString(s, this.config)).join(",\n");
+      rule.selector = parsedSelectors.map(s => block.rewriteSelectorToString(s, this.config, analyzer ? analyzer.reservedClassNames() : new Set())).join(",\n");
     });
 
-    resolver.resolve(root, block);
+    resolver.resolve(root, block, analyzer ? analyzer.reservedClassNames() : new Set());
```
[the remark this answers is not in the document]
ramitha: Fixed for now and we can get rid of it after we start precomputing the classnames

packages/@css-blocks/core/src/Analyzer/Analysis.ts
```diff
@@ -325,7 +338,7 @@ export class Analysis<K extends keyof TemplateTypes> {
    * @param options The plugin options that are used to parse the blocks.
    * @param postcssImpl The instance of postcss that should be used to parse the block's css.
    */
-  static async deserialize (
+  async deserialize (
```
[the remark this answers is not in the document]
ramitha: For now, I've changed it so it works it gets it from the `parent`. Will revisit with the refactor.

PR #347, opened 2019-11-05
Import autocomplete

jackson-dean:


packages/@css-blocks/language-server/src/util/blockUtils.ts
```diff
@@ -29,3 +35,88 @@ export async function parseBlockErrors(parser: BlockParser, blockFsPath: string,
...
+    return completionItems;
+  }
+
+  let blockDirPath = path.dirname(URI.parse(documentUri).fsPath);
+  let absoluteImportPath = path.resolve(blockDirPath, relativeImportPath);
+  let globPatternSuffix = relativeImportPath.endsWith("/") ? "/*" : "*";
+  let currentFileExtension = path.extname(documentUri);
+
+  let pathNames: string[] = await new Promise(r => {
+    glob(`${absoluteImportPath}${globPatternSuffix}`, async (_, paths) => r(paths));
```
chriseppstein: glob takes an option `cwd` that is the starting directory for matching the glob pattern. Did you consider using that option here for constructing the directory? I have an aversion to passing user input into glob without sanitizing it... the user isn't typing a glob they are typing a filename which might contain characters that are special chars in the glob syntax.

packages/@css-blocks/language-server/src/test/fixtures/ember-classic/styles/components/import-completions.block.css
```diff
@@ -0,0 +1,2 @@
+@block randoBlock from "../";
+@export utils from "../blocks/";
```
chriseppstein: I think we need three more test cases (note, I'm using `|` to indicate the location of the cursor in the examples below):
1. for nested subdirectories E.g. (`@block myBock from "dir/subdir/|"`)
2. for a half-written filename E.g. (`@block anotherBlock from "foo/another-|"`)
3. for local directory references. E.g. (`@block relativeBlock from "./|"`)

packages/@css-blocks/language-server/src/documentLinksProviders/blockLinkProvider.ts
```diff
@@ -5,7 +5,7 @@ import { URI } from "vscode-uri";
 
 import { isBlockFile } from "../util/blockUtils";
 
-const LINK_REGEX = /from\s+(['"])([^'"]+)\1;/;
+export const LINK_REGEX = /from\s+(['"])([^'"]+)\1;?/;
```
chriseppstein: Are we sure we want to enforce the presence of the closing quote when suggesting the contents for the quote? Some users may not have the feature enabled that automatically inserts the closing quote mark when the type the opening quote mark.

packages/@css-blocks/language-server/src/util/blockUtils.ts
chriseppstein: Also if we're ultimately just listing the contents of a directory and filtering filenames by `.startsWith`, I feel like `fs.readdir` is just as simple to use here and avoids a complex dependency.

packages/@css-blocks/language-server/src/util/blockUtils.ts
```diff
@@ -29,3 +35,88 @@ export async function parseBlockErrors(parser: BlockParser, blockFsPath: string,
...
+  });
+
+  let fileInfos: PathCompletionCandidateInfo[] = await Promise.all(pathNames.map(pathName => {
+    return new Promise(r => {
+      fs.stat(pathName, (_, stats) => r({ pathName, stats }));
```
chriseppstein: stat can have an error... I think we need to handle it and ignore it safely.

packages/@css-blocks/language-server/src/documentLinksProviders/blockLinkProvider.ts
jackson-dean: That's a good point. I opted to just have a regex specific to this use case which is a bit looser rather than to keep relaxing this one to make it fit both use cases.

packages/@css-blocks/language-server/src/test/fixtures/ember-classic/styles/components/import-completions.block.css
jackson-dean: 👍 I added these cases.

packages/@css-blocks/language-server/src/util/blockUtils.ts
```diff
@@ -29,3 +35,88 @@ export async function parseBlockErrors(parser: BlockParser, blockFsPath: string,
...
+  pathName: string;
+  stats: fs.Stats;
+}
+
+function maybeCompletionItem(completionCandidateInfo: PathCompletionCandidateInfo, currentFileExtension: string): CompletionItem | null {
+    let completionKind: CompletionItemKind | undefined;
+
+    if (completionCandidateInfo.stats.isDirectory()) {
+      completionKind = CompletionItemKind.Folder;
+    } else if (completionCandidateInfo.stats.isFile() && completionCandidateInfo.pathName.endsWith(`.block${currentFileExtension}`)) {
```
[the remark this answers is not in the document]
jackson-dean: ahh, nice. I updated to just check for `.block`

packages/@css-blocks/language-server/src/util/blockUtils.ts
jackson-dean: gotcha. yeah, i was naively using glob to filter the results, but that is a good point if there are special chars in the file name it becomes a glob pattern, etc. On second look it is totally unnecessary to use glob for this since the vscode auto completion UI already does filtering for us. We just need to send whatever is in the directory and as the user types, vscode will whittle down the list in the client editor UI.

packages/@css-blocks/language-server/src/util/blockUtils.ts
jackson-dean: gotcha. i changed to allow this to resolve to `null` if stats object is not falsy.

packages/@css-blocks/language-server/src/util/blockUtils.ts
```diff
@@ -81,37 +80,43 @@ async function getImportPathCompletions(documentUri: string, relativeImportPath:
...
-  let absoluteImportPath = path.resolve(blockDirPath, relativeImportPath);
-  let globPatternSuffix = relativeImportPath.endsWith("/") ? "/*" : "*";
-  let currentFileExtension = path.extname(documentUri);
-
+  let relativeScanDir = relativeImportPath.endsWith(path.sep) ? relativeImportPath : relativeImportPath.substring(0, relativeImportPath.lastIndexOf(path.sep) + 1);
+  let absoluteScanDir = path.resolve(blockDirPath, relativeScanDir);
+  let blockFsPath = URI.parse(documentUri).fsPath;
   let pathNames: string[] = await new Promise(r => {
-    glob(`${absoluteImportPath}${globPatternSuffix}`, async (_, paths) => r(paths));
+    fs.readdir(absoluteScanDir, (_, paths) => r(paths || []));
```
chriseppstein: readdir can have an error if it doesn't exist or the permissions are wonky. we should handle that here.

packages/@css-blocks/language-server/src/util/blockUtils.ts
```diff
@@ -87,16 +87,18 @@ async function getImportPathCompletions(documentUri: string, relativeImportPath:
     fs.readdir(absoluteScanDir, (_, paths) => r(paths || []));
   });
 
-  let fileInfos: PathCompletionCandidateInfo[] = await Promise.all(pathNames.map(pathName => {
+  let fileInfos: (PathCompletionCandidateInfo | null)[] = await Promise.all(pathNames.map(pathName => {
     return new Promise(r => {
       let absolutePath = `${absoluteScanDir}/${pathName}`;
-      fs.stat(absolutePath, (_, stats) => r({ pathName: absolutePath, stats }));
+      fs.stat(absolutePath, (_, stats) => r(
+        stats ? { pathName: absolutePath, stats } : null,
```
chriseppstein: The callback's first argument is an error if there's an error. That's what we should be checking. https://nodejs.org/api/fs.html#fs_fs_stat_path_options_callback

Alternatively, we can use the promisified version with an await and just try/catch it. https://nodejs.org/api/fs.html#fs_fspromises_stat_path_options

packages/@css-blocks/language-server/src/util/blockUtils.ts
chriseppstein: the promise version of readdir with await and a try/catch is also a good way to go: https://nodejs.org/api/fs.html#fs_fspromises_readdir_path_options

jackson-dean: per our discussion, the promise apis are not yet stable in the node version used by vscode. However, I updated the error handling to be much more explicit.

packages/@css-blocks/language-server/src/util/blockUtils.ts
```diff
@@ -29,3 +34,124 @@ export async function parseBlockErrors(parser: BlockParser, blockFsPath: string,
...
+  try {
+    return await new Promise((resolve, reject) => {
+      fs.stat(filePath, (error, stats) => {
+        error ? reject(error) : resolve(stats);
+      });
+    });
+  } catch (e) {
+    return null;
+  }
+}
```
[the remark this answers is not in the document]
chriseppstein: There's a similar refactor of `getFilesInDirectory()`.

packages/@css-blocks/language-server/package.json
```diff
@@ -6,7 +6,7 @@
   "author": "LinkedIn Corporation",
   "license": "BSD-2-Clause",
   "engines": {
-    "node": "*"
+    "node": "10.11.0"
```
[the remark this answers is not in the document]
chriseppstein: This is making our CI build fail where we test against node 8, 10, and 12: 

```
error @css-blocks/language-server@0.0.1: The engine "node" is incompatible with this module. Expected version "10.11.0". Got "8.16.2"
```

I think it's ok to leave the node engine version as `*` for this field for now. If we decide to take advantage of some new node features in the future we can adjust the build to skip VS code related packages for come of the node versions we test against.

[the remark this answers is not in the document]
jackson-dean: It is fine with me to leave it as `*` as long as it's generally understood that it has to track whatever vscode is built with.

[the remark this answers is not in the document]
jackson-dean: I changed it back for now.

packages/@css-blocks/language-server/src/util/blockUtils.ts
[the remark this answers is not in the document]
jackson-dean: sure, it works for me 👍

PR #348, opened 2019-11-08
Vs code importer

ramitha:


packages/@css-blocks/language-server/src/Importer.ts
```diff
@@ -0,0 +1,32 @@
...
+export class LSImporter extends NodeJsImporter {
+  documents: TextDocuments;
+  constructor(documents: TextDocuments) {
+    super();
+    this.documents = documents;
+  }
+
+  async import(identifier: string, config: Readonly<import("@css-blocks/core/dist/src").Configuration>): Promise<import("@css-blocks/core/dist/src").ImportedFile> {
+    // the uri expected is that of a file
+    let clientDocument = this.documents.get(`file://${identifier}`);
```
jackson-dean: You can also use the `URI` utility class to convert the identifier to a proper uri:
```
import { URI } from "vscode-uri";
...
URI.file(indentifier).toString();
```
You can search for `URI.file` in the project to some other examples.

packages/@css-blocks/language-server/src/run.ts
```diff
@@ -11,6 +12,6 @@ const pathTransformer = new EmberClassicTransformer(Syntax.css);
 const connection = createConnection(ProposedFeatures.all);
 const documents = new TextDocuments();
 
-let server = new Server(connection, documents, pathTransformer);
+let server = new Server(connection, documents, pathTransformer, {importer: new LSImporter(documents)});
```
jackson-dean: I am probably just missing something or there is just a gap in my understanding about how this works, but I'm having trouble finding where and how this importer is actually used? I can't find any reference to it in the server class itself. I was also curious why it is passed in an options hash?

packages/@css-blocks/language-server/README.md
```diff
@@ -0,0 +1,23 @@
...
+  2. choose “Open folder”
+  3. navigate to css-blocks and choose to open the css-blocks.code-workspace
+     file directly
+3. Open the debug panel by pressing `cmd+shift+D`
+4. At the top there should be a drop down. select Client + Server
+5. After that, pressing f5 or clicking the “play” icon next to the selected
+   configuration should run the extension
+6. If youre making changes to the server and you need to compile them as the
+   changes are made, go to `css-blocks/packages/@css-blocks/language-server` and
+   run `yarn watch`
```
chriseppstein: Thank you for adding this!

packages/@css-blocks/core/src/configuration/types.ts
```diff
@@ -20,8 +20,8 @@ export interface Configuration {
   preprocessors: Preprocessors;
 
   /**
-   * An importer is an object that is in charge of findi
-   *
+   * An importer is an object that is in charge of finding the path to the block
+   * file from it's @block directive
```
chriseppstein: Good catch! Although, technically the importer is in charge of finding the contents of a block from its `@block` directive.

grammar nit: `its` not `it's`

packages/@css-blocks/language-server/package.json
```diff
@@ -32,6 +32,6 @@
     "posttest": "yarn run lint",
     "test": "mocha dist/test --opts src/test/mocha.opts",
     "test:runner": "mocha dist/test",
-    "watch": "watch 'yarn run test' './src' --wait=3"
+    "watch": "watch 'yarn run compile' './src' --wait=3"
```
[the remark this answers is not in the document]
ramitha: Just that I was a little impatient. Will change it back :)

packages/@css-blocks/language-server/src/run.ts
ramitha: The blockFactory's methods use the importer to read the contents of any block. Be default, it uses the nodeJSImporter, but it's configurable and we can specify an importer as a config when creating the blockFactory. 

https://github.com/linkedin/css-blocks/blob/7e030fb95b73f6838b0af8f78f42d81ba0566023/packages/%40css-blocks/core/src/BlockParser/BlockFactory.ts#L51

packages/@css-blocks/core/src/configuration/types.ts
ramitha: I've been making it's errors more and more when I type!

PR #351, opened 2019-11-14
Configuration file support for CLI and VSCode

chriseppstein:


packages/@css-blocks/language-server/src/Importer.ts
```diff
@@ -7,16 +7,18 @@ import { URI } from "vscode-uri";
  * already opened on the client. Otherwise, it proxies to the NodeJSImporter to
  * read the file from the disk
  */
-export class LSImporter extends NodeJsImporter {
+export class LSImporter implements Importer {
+  baseImporter: Importer;
   documents: TextDocuments;
-  constructor(documents: TextDocuments) {
-    super();
```
jackson-dean: Is the super call removed intentionally here?

packages/@css-blocks/language-server/src/Server.ts
```diff
@@ -1,44 +1,53 @@
...
-    this.blockParser = createBlockParser(config, blockFactory);
-
     this.registerDocumentEvents();
     this.registerConnectionEvents();
   }
 
+  get config(): Readonly<Configuration> {
+    if (!this._config) {
+      this._config = resolveConfiguration({});
+      return this._config;
```
jackson-dean: nit: it doesn't look like the early return is needed

packages/@css-blocks/language-server/src/Server.ts
```diff
@@ -91,8 +100,35 @@ export class Server {
...
+    // TODO #1: But the rest of the config lookup logic should remain basically the same.
+    // TODO #2: There should be a configuration option to set the rootDir for CSS Blocks
+    // TODO #2: as well as a configuration option to set the configuration file explicitly
+    // TODO #2: instead of only doing a search for the configuration file.
+    let result: Partial<Configuration> | null = null;
+    let rootDir: string | null = null;
+    if (params.workspaceFolders) {
+      for (let wsFolder of params.workspaceFolders) {
+        let folderPath = URI.parse(wsFolder.uri).fsPath;
+        result = await searchForConfig(folderPath);
```
jackson-dean: i'm not sure it matters all that much, but do the searches need to happen in sequence? Would it be faster to create an array of promises and do Promise.all and filter it rather than awaiting in each iteration of the loop?

chriseppstein: This is just placeholder code until we can spin up a language server per workspace; not worth optimizing, imo.

packages/@css-blocks/language-server/src/Server.ts
chriseppstein: Removed. I thought it was needed to make the typechecker happy but TS has gotten smarter in recent releases.

packages/@css-blocks/language-server/src/Importer.ts
chriseppstein: Yes, the Importer is no longer a subclass of the `NodeJSImporter` (or any importer). It just implements the `Importer` interface and wraps calls to an arbitrary importer instance that is passed in at the time of construction. This was necessary to support the possibility that a customized importer was specified in the configuration file.

packages/@css-blocks/language-server/src/Server.ts
chriseppstein: 5f12eb5405c2f8c22c7ecca7cb30398d82131e17

packages/@css-blocks/language-server/src/Importer.ts
jackson-dean: ahh, yes I see it now. I overlooked that extends -> implements change 👍

packages/@css-blocks/language-server/src/Server.ts
jackson-dean: 👍 sgtm!

PR #352, opened 2019-11-15
feat: Basic preprocessor support.

chriseppstein:
This PR adds basic support for any preprocessors that are set up in the css-blocks' configuration file. Because any extension is technically supported, we have to go to the disk to see what files actually exist according to expected naming conventions so that we can return the path to the correct file. Stylesheets that appear to be block files (e.g. named `*.block.*`) but aren't supported by the preprocessor configuration are basically ignored right now because otherwise CSS Blocks would just throw an error if we tried to parse it.

TODO:
- [x] Add Tests


jackson-dean:
implementation lgtm!

PR #378, opened 2020-01-02
Update package.json

lucacorti:
Test folders are not there and ember-cli@3.15.1 fails silently when checking the addon.


amiller-gh:
Mmm, yeah if I remember correctly these paths are required for us to run production-like tests in the addon. Its a bit of a fun setup because of the pretty deep Ember CLI integration. You're right, it'd be better if these weren't here polluting the package.json, but just removing them won't do. 

If https://github.com/ember-cli/ember-cli/pull/8907 does fix the issue, you should be un-blocked. But, we may still want to consider either a) stripping these lines out before package publish, or b) moving integration tests out to another package in this monorepo (a is quick and easy, b is more involved but probably a better long-term solution)

lucacorti:
@amiller-gh I'm on ember-cli 3.15.1 and seeing this, so I guess ember-cli/ember-cli#8907 is not in it or not working for me.

lucacorti:
ember-cli 3.15.2 fixes this, just emits a warning:

```
WARNING: The 'package.json' file for the addon at /Users/luca/Desktop/dottori/node_modules/@css-blocks/ember-cli
  specifies invalid, malformed or missing addons at relative paths
    'tests/dummy/lib/in-repo-addon'
    'tests/dummy/lib/in-repo-engine'
    'tests/dummy/lib/in-repo-lazy-engine'
```

PR #387, opened 2020-01-23
feat: WIP for find all references.

chriseppstein:


packages/@css-blocks/language-server/src/Server.ts
```diff
@@ -115,13 +117,51 @@ export class Server {
...
+        });
+
+        const rootDir = "/Users/jcdean/ws/css-blocks-demo/app";
+        const templates = glob.sync("**/*.hbs", { ignore: ["node_modules"], cwd: rootDir });
+        let itemAtCursor = getItemAtCursor(document.getText(), params.position);
+        let references: Location[] = [];
+        let analyses;
+
+        try {
+          analyses = await analyzer.analyze(rootDir, templates);
```
chriseppstein: We have a method on `ElementAnalyzer` called `analyzeForRewrite`

https://github.com/linkedin/css-blocks/blob/master/packages/%40css-blocks/glimmer/src/ElementAnalyzer.ts#L65 

This method returns an ElementAnalysis that has associations to the template AST nodes at least for dynamic styles. I think this will allow you to connect the dots better... we might need to track nodes for the static styles too.

packages/@css-blocks/language-server/src/Server.ts
```diff
@@ -115,13 +117,51 @@ export class Server {
...
+        } catch (e) {
+          console.log(e);
+          return references;
+        }
+
+        const serialized = analyses.serialize();
```
chriseppstein: As you mentioned in chat, I think you want to do this without serializing first. Especially given my previous comment about the analyzeForRewrite.

chriseppstein:
In ElementAnalyzer, I think we can get rid of the distinction of `analyze` and `analyzeForRewrite` and just make everything use `analyzeForRewrite` (which we would then rename to `analyze`). There's very little benefit to this, it's really just a microoptimization to store less data in situation where we don't need it, but it's adding a lot of unnecessary code complexity.

Then we need to augment how much information we store about where the styles are in the template. The current source locations that we capture aren't specific enough for vscode. For this, we need to augment `ElementAnalysis` in `@css-blocks/core` so that for every static and dynamic style we capture there's an associated node location. For instance I suspect we'll need to change

```ts
interface TrueCondition<Container extends BlockClass | number = BlockClass> {
  whenTrue: Array<Container>;
}
```

to be 

```ts
interface StyleAndLocation<Style> {
  style: Style;
  location: SourceLocation;
}
interface TrueCondition<Container extends BlockClass | number = BlockClass> {
  whenTrue: Array<StyleAndLocation<Container>>;
}
```

PR #394, opened 2020-02-13
chore: Refactor to use TypeScript & lerna builds.

abritinthebay:
This is a first pass, with obvious improvements that can be made, but it's better than not being in TS.

* Set up project to extend monorepo's `tsconfig.json`
* Updated monorepo config to fix issues seen during conversion
* Added build and compile scripts
* Added ember-cli-typescript to allow (non app) code to be Typescript seamlessly
* Added TS-Node to allow TypeScript in code included in `index.js` for local development
* Added global (to addon) `.d.ts` file for the types needed (inc mocking up some of Ember's types)
* Pulled apart code into separate files for easier maintenance and reading
* Added a lot of guard logic
* Removed `lodash.merge` & replaced with ESNext/TS idioms


abritinthebay:
Looks like playground is running into issues on CI with the new setup. Will noodle on how to fix that.

abritinthebay:
Fixed: needed to explicitly pass `ts-node` the tsconfig.json file, else it gets lost in the link/build process.

packages/@css-blocks/ember-cli/index.js
```diff
@@ -1,419 +1,47 @@
...
+ * a `lerna` or `yarn` step with `run compile` the lib will use the
+ * transpiled version. So changes won't be reflected until you clean it up.
+ */
+const shouldRunProdBuild = fs.existsSync(`${__dirname}/lib/addon.js`);
+// disable eslint here as the transpiled addon may not exist, and that's ok.
+/* eslint-disable-next-line node/no-missing-require */
+const prodBuild = () => require('./lib/addon').default; // ES Module
+
+const devBuild = () => {
+  console.log("[@css-blocks/ember-cli] Starting Dev Build & transpiling with TS Node...");
```
abritinthebay: Note: if we want to do this not dynamically (in a prebuild step..?) that's fine, but I ran into some odd issues/difficulties doing that. Possibly just due to my inexperience with TS tho.

abritinthebay:
Was feeling curious so cleaned this up a bunch based on what we discussed. There are still some custom types but they're explicitly imported (from `_utils`) and they may be able to be trimmed some more.

PR #399, opened 2020-02-19
More robust importing

chriseppstein:
My fix for https://github.com/linkedin/css-blocks/issues/248 introduced a stack on error objects that allowed errors to track the location of the error within a series of imports. This turned out to be problematic because our blocks are singletons and so the errors for that block also want to be singletons. Because a block might be imported (and exported) from different blocks, the stack needed to be tracked differently.

This PR adds the CascadingError error for indicating an error was caused by some previous error and uses it for block imports and exports.

This allowed me to stop doing weird things trying to parse a block a second time if it previously had an error -- just to get a different error instance.

I've done a pretty major refactor of how `@block` at-rules are resolved. The old code was a convoluted mess of promise chains spawned while AST walking and then resolved afterwards. I've rewritten with async/await and have separated out the AST walking from the work done based on what we found in the document.

Building on the MultipleBlockErrors type by @ramitha, I've updated our error reporting in a few places to describe the sequence of events that lead up to an error.

For example, the CLI now displays the root cause for an error like so:

```
error   test/fixtures/basic/error.block.css
        Two distinct classes cannot be selected on the same element: .foo.bar
        At test/fixtures/basic/error.block.css:1:5
        1: .foo.bar {
        2:   color: red;
caused  test/fixtures/basic/transitive-error.block.css
        Error in imported block.
        At test/fixtures/basic/transitive-error.block.css:1:1
        1: @block error from "./error.block.css";
        2:

error   test/fixtures/basic/transitive-error.block.css
        No Block named "error" found in scope.
        At test/fixtures/basic/transitive-error.block.css:4:3
        3: :scope {
        4:   extends: error;
        5: }

Found 2 errors in 1 file.
```

This PR also introduces a utility function `allDone` that is like `Promise.all()` except it waits for all the work to complete first. In most places, I've stopped using `Promise.all()` after I became aware that it has potential race condition behavior within a broccoli build/rebuild loop.


packages/@css-blocks/cli/test/cli-test.ts
```diff
@@ -43,18 +43,26 @@ Found 1 error in 1 file.
     let cli = new CLI();
     await cli.run(["validate", fixture("basic/transitive-error.block.css")]);
 
-    assert.equal(cli.output,
-                 `error\t${relFixture("basic/transitive-error.block.css")}
+    assert.equal(
+      cli.output,
+      `error\ttest/fixtures/basic/error.block.css
 \tTwo distinct classes cannot be selected on the same element: .foo.bar
```
abritinthebay: Just curious: why tab literals in a template string?

packages/@css-blocks/core/src/BlockParser/features/export-blocks.ts
```diff
@@ -50,61 +51,70 @@ export async function exportBlocks(block: Block, factory: BlockFactory, file: st
...
               block.addError(new errors.InvalidBlockSyntax(
-                `Unnecessary re-export of default Block.`,
+                `Cannot have duplicate Block export of same name: "${remoteName}".`,
                 sourceRange(factory.configuration, block.stylesheet, file, atRule),
               ));
-            }
+            } else {
+              let localName = blockNames[remoteName];
+              if (!CLASS_NAME_IDENT.test(localName)) {
+                block.addError(new errors.InvalidBlockSyntax(
```
abritinthebay: Might just be me, but all these cases feel super abstract-able into some kind of `addBlockError(block: Block, msg: string): void` type of signature. 

Would make this easier to read (for me, anyhow)

packages/@css-blocks/cli/test/cli-test.ts
chriseppstein: Not anything super important but I think the presence of the tab stop might be missed so I wanted to make it more clear what was going on. Also, this is more robust against people who like to switch to tabs while they're editing a file and then put it back to spaces later (Good IDEs wouldn't mess this up, but some ways of doing this with vim would, for example.)

packages/@css-blocks/core/src/BlockParser/features/export-blocks.ts
chriseppstein: yeah, in other places, we've abstracted out the adding/raising of errors within a particular context and it definitely cleans things up. Let's iterate on that in a separate PR though.

abritinthebay: Yeah, didn't want to raise it as an issue, just a comment :D

PR #402, opened 2020-02-21
feat(eyeglass): Adds new package that enables simple Eyeglass support.

### LINKED ISSUE #398: Productize Eyeglass support
Right now we're doing this in `css-blocks.config.js` to add eyeglass support to a project:

```js
const sass = require("node-sass");
const eyeglass = require("eyeglass");
function scss(file, data, _configuration, _sourceMap) {
  return new Promise((resolve, reject) => {
    const sassOptions = {
      file,
      data,
      outputStyle: "expanded",
      sourceMap: true,
      outFile: file.replace(/scss$/, "css"),
    };
    sass.render(eyeglass(sassOptions), (err, res) => {
      if (err) {
        reject(err)
      } else {
        resolve({
          content: res.css.toString(),
          sourceMap: res.map.toString(),
          dependencies: res.includedFiles,
        })
      }
    });
  })
}
module.exports = {
  preprocessors: { scss }
};
```

Instead we should be able to do this:

```js
const sass = require("node-sass");
const eyeglass = require("eyeglass");
const eyeglassAdaptor = require("@css-blocks/eyeglass");

const sassOptions = {
  outputStyle: "expanded",
};

const scss = eyeglassAdaptor(sass, eyeglass, sassOptions);

module.exports = {
  preprocessors: { scss }
};
```

A few things to note here: Sass and Eyeglass are dependency injected. I don't want us to manage what versions or implementations get installed.

The options other than `outputStyle` that we're setting need to be set that way, we shouldn't allow them to be overridden by the caller so we should `Object.assign` into their options, not the other way around.

abritinthebay:
Fixes #398 

Should be a cleaner TS implementation than my previous PR (thanks for the feedback and help using Namespaces as types). I'm using Sinon spies in the tests, and because of that found I didn't need to use the module importer mock you mentioned (as rather than testing all of css-blocks in the unit tests I'm just testing what _this package_  takes in and puts out, as well as that the created functions are called). 

Code test coverage is 100%, which is nice.

### DIFF
--- a/packages/@css-blocks/eyeglass/package.json
+++ b/packages/@css-blocks/eyeglass/package.json
@@ -1,0 +1,61 @@
+{
+  "name": "@css-blocks/eyeglass",
+  "author": "Gregory Wild-Smith <gwildsmith@linkedin.com>",
+  "description": "Tool to provide easy interoperation with the Eyeglass SASS module manager",
+  "license": "BSD-2-Clause",
+  "version": "1.0.0-alpha.7",
+  "main": "dist/src/index.js",
+  "types": "dist/src/index.d.ts",
+  "files": [
+    "bin",
+    "dist",
+    "src",
+    "README.md",
+    "CHANGELOG.md"
+  ],
+  "readme": "README.md",
+  "keywords": [
+    "eyeglass",
+    "css blocks",
+    "css",
+    "blocks"
+  ],
+  "scripts": {
+    "test": "yarn run test:runner",
+    "test:runner": "mocha --opts test/mocha.opts dist/test",
+    "compile": "tsc --build",
+    "pretest": "yarn run compile",
+    "posttest": "yarn run lintfix",
+    "prepublish": "rm -rf dist && yarn run compile && yarn run lintall",
+    "lint": "tslint -t msbuild --project . -c tslint.cli.json",
+    "lintall": "tslint -t msbuild --project . -c tslint.release.json",
+    "lintfix": "tslint -t msbuild --project . -c tslint.cli.json --fix",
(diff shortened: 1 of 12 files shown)


packages/@css-blocks/eyeglass/src/index.ts
```diff
@@ -0,0 +1,29 @@
+import { ProcessedFile } from "@css-blocks/core";
+import Eyeglass from "eyeglass"; // works, even tho a cjs export. huh.
+import { Options, Result, SassError } from "node-sass";
+import SassImplementation from "node-sass";
+
+export function adaptor(foo: typeof SassImplementation, eyeglass: typeof Eyeglass, options: Options = {}) {
```
chriseppstein: let's call it `sass` instead of `foo`.

packages/@css-blocks/eyeglass/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@css-blocks/eyeglass",
+  "author": "Gregory Wild-Smith <gwildsmith@linkedin.com>",
+  "description": "Tool to provide easy interoperation with the Eyeglass SASS module manager",
+  "license": "BSD-2-Clause",
+  "version": "1.0.0-alpha.6",
```
chriseppstein: we're at alpha.7 now.

packages/@css-blocks/eyeglass/src/index.ts
abritinthebay: ... argh. Yes 🤦‍♂

abritinthebay:
Updated with those couple of small changes 😄

PR #403, opened 2020-02-25
feat(style-of): Allows positional arguments to be passed.

### LINKED ISSUE #395: feat: style-of helper should allow positional arguments to be passed.
right now we require the hash style arguments even for things like `block:scope` which are just set to `true`. Because helpers can take positional arguments, we can also accept `block:scope` without a "value" assigned to it.

abritinthebay:
Fixes #395

### DIFF
--- a/packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
+++ b/packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
@@ -16,10 +16,15 @@ import { GlimmerAnalysis } from "./Analyzer";
 import { getEmberBuiltInStates, isEmberBuiltIn } from "./EmberBuiltins";
 import { ResolvedFile } from "./Template";
 import {
+  AnalyzableNode,
+  AnalyzableProperty,
   cssBlockError,
+  isAnalyzableProperty,
+  isAttrNode,
   isBooleanLiteral,
   isConcatStatement,
   isElementNode,
+  isHashPair,
   isMustacheStatement,
   isNullLiteral,
   isNumberLiteral,
@@ -45,8 +50,6 @@ const DEFAULT_BLOCK_NAME = "default";
 const DEFAULT_BLOCK_NS = "block";
 
 const debug = debugGenerator("css-blocks:glimmer:element-analyzer");
-
-type AnalyzableNode = AST.ElementNode | AST.BlockStatement | AST.MustacheStatement | AST.SubExpression;
 
 export function isStyleOfHelper(node: AnalyzableNode): node is AST.MustacheStatement | AST.SubExpression {
   if (!(isMustacheStatement(node) || isSubExpression(node))) return false;
@@ -133,19 +136,51 @@ export class ElementAnalyzer {
     }
   }
 
-  *eachAnalyzedAttribute(node: AnalyzableNode): Iterable<[string, string, AST.AttrNode | AST.HashPair]> {
-    if (isElementNode(node)) {
-      for (let attribute of node.attributes) {
-        let [namespace, attrName] = this.isAttributeAnalyzed(attribute.name);
-        if (namespace && attrName) {
-          yield [namespace, attrName, attribute];
-        }
-      }
+  private _throwIfInvalidAttribute(node: AnalyzableNode, property: AnalyzableProperty, name: string): void {
+    if (isPathExpression(property) && name === "class") {
+      const name = this._getAnalyzableAttributeName(property);
+      throw cssBlockError(`The ${name} attribute must contain a value and is not allowed to be purely positional. Did you mean ${name}="foo"?`, node, this.template);
+    }
+  }
+
+  private _getAnalyzableAttributeName(attribute: AnalyzableProperty): string | void {
+    if (isAttrNode(attribute)) {
+      return attribute.name;
+    } else if (isHashPair(attribute)) {
+      return attribute.key;
+    } else if (isPathExpression(attribute)) {
+      return attribute.original;
     } else {
-      for (let pair of node.hash.pairs) {
-        let [namespace, attrName] = this.isAttributeAnalyzed(pair.key);
-        if (namespace && attrName) {
-          yield [namespace, attrName, pair];
+      assertNever(attribute);
+    }
+  }
+
+  *eachAnalyzedAttribute(node: AnalyzableNode): Iterable<[string, string, AnalyzableProperty]> {
+    // Intital list may also contain general Expressions, we filter that later
+    const propertyList: (AnalyzableProperty[] | AST.Expression[])[] = [];
+    // set up the list of attributes (or multiple lists!) that we want to check
+    // attributes for Element nodes, hash pairs and params for Handlebars nodes.
+    if (isElementNode(node)) {
+      propertyList.push(node.attributes);
+    } else {
+      propertyList.push(node.params, node.hash.pairs); // listed in order they'd need to be in Handlebars code
+    }
+    // looping through like this means the lists will be checked in order - eg, all hash pairs then all params
+    // this isn't strictly needed, but helps with debugging.
+    for (let list of propertyList) {
+      // get each attribute, check it is analyzable, yield it if it parses.
+      // will throw if the attribute name is not allowed
+      // (ie - it's a positional prop but attribute requires data, like block:class)
+      for (let property of list) {
+        if (isAnalyzableProperty(property)) {
+          const name = this._getAnalyzableAttributeName(property);
+          if (name) {
+            const [namespace, attrName] = this.isAttributeAnalyzed(name);
+            if (namespace && attrName) {
+              this._throwIfInvalidAttribute(node, property, attrName);
+              yield [namespace, attrName, property];
+            }
+          }
         }
       }
     }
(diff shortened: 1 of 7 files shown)


packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -133,19 +137,43 @@ export class ElementAnalyzer {
     }
   }
 
-  *eachAnalyzedAttribute(node: AnalyzableNode): Iterable<[string, string, AST.AttrNode | AST.HashPair]> {
+  private _throwIfInvalidAttribute(node: AnalyzableNode, property: AnalyzableProperty, name: string): void {
+    if (isPathExpression(property) && name === "class") {
+      throw cssBlockError(`The namespace:class attribute must contain a value and is forbidden to be positional. Did you mean block:class="foo"?`, node, this.template);
```
chriseppstein: We should know what the value for `namespace` is. Can we put that in this error message so people can sort out the bad attribute faster?

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -133,19 +137,43 @@ export class ElementAnalyzer {
...
+    }
+  }
+
+  private _getAnalyzableAttributeName(attribute: AnalyzableProperty): string {
+    return isAttrNode(attribute) ? attribute.name :
+            isHashPair(attribute) ? attribute.key :
+              isPathExpression(attribute) ? attribute.original :
+                "";
```
chriseppstein: This is a case where `assertNever()` is ideal. Because the type guards are exhaustive, the empty string case should never happen. And if we did add more types to the `AnalyzableProperty` alias, we'd want the compiler to tell us about this code so we can add custom logic for it.

```suggestion
    if (isAttrNode(attribute)) {
      return attribute.name;
    } else if (isHashPair(attribute)) {
      return attribute.key;
    } else if (isPathExpression(attribute)) {
      return attribute.original;
    } else {
      assertNever(attribute);
    }
```

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -251,13 +279,13 @@ export class ElementAnalyzer {
...
-  private processClass(namespace: string, node: AST.AttrNode | AST.HashPair, element: TemplateElement, forRewrite: boolean): void {
+  private processClass(namespace: string, node: AnalyzableProperty, element: TemplateElement, forRewrite: boolean): void {
     let statements: AST.Node[];
 
-    let value = node.value;
-
-    if (isConcatStatement(value)) {
-      statements = value.parts;
+    if (isPathExpression(node)) {
+      statements = [node];
```
chriseppstein: We have an assertion elsewhere that says that classes without values aren't allowed so I don't think this AST node is ever passed to this function. The `statements` in this case are supposed to be the value of the class, not the attribute name.

In this case, I'd set the parameter `node` to have a type of `Exclude<AnalyzableProperty, AST.PathExpression>` because the calling code is in error if it were to want to pass that value.

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -361,43 +392,45 @@ export class ElementAnalyzer {
     }
     let staticSubStateName: string | undefined = undefined;
     let dynamicSubState: AST.MustacheStatement | AST.ConcatStatement | AST.SubExpression | AST.PathExpression | undefined = undefined;
-    let value = node.value;
-    if (isTextNode(value)) {
-      staticSubStateName = value.chars;
+
+    if (isPathExpression(node) || isSubExpression(node)) {
+      // treat it like a boolean value true and do nothing
```
chriseppstein: I don't understand the `isSubExpression(node)` case. I don't think that should ever happen. It looks like the type for `node` doesn't include SubExpression. I think we can just remove that from this line and nothing will break.

packages/@css-blocks/glimmer/src/utils.ts
```diff
@@ -72,3 +72,12 @@ export function isUndefinedLiteral(value: AST.Node | undefined): value is AST.Un
 export function isPathExpression(value: AST.Node | undefined): value is AST.PathExpression {
   return !!value && value.type === "PathExpression";
 }
+export function isHashPair(value: AST.Node | undefined): value is AST.HashPair {
+  return !!value && value.type === "HashPair";
+}
+export function isAttrNode(value: AST.Node | undefined): value is AST.AttrNode {
+  return !!value && value.type === "AttrNode";
+}
+export function isAnalyzableProperty(value: AST.Node | undefined): value is (AST.AttrNode | AST.HashPair | AST.PathExpression) {
```
chriseppstein: I think the return type here should be `AnalyzableProperty`, was there a reason why you wanted to enumerate those types explicitly here?

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
abritinthebay: oh good point! Will do.

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
abritinthebay: It's not actually exhaustive. The "" will be for any attribute that isn't one of those. Like anything on a html element for example. Or any non-supported attribute.

I actually use this in the other PR for noting the unsupported attributes.

packages/@css-blocks/glimmer/src/utils.ts
abritinthebay: Nope, was just done before I created that type so must have left it at that. Good catch.

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
abritinthebay: I don't remember why I added that... probably was some type issue earlier and then I fixed it but forgot about this test... will remove

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
abritinthebay: Updated!

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -133,19 +136,44 @@ export class ElementAnalyzer {
...
+      const name = this._getAnalyzableAttributeName(property);
+      throw cssBlockError(`The ${name} attribute must contain a value and is not allowed to be purely positional. Did you mean ${name}="foo"?`, node, this.template);
+    }
+  }
+
+  private _getAnalyzableAttributeName(attribute: AnalyzableProperty): string {
+    return isAttrNode(attribute) ? attribute.name :
+            isHashPair(attribute) ? attribute.key :
+              isPathExpression(attribute) ? attribute.original :
+                assertNever(attribute);
```
chriseppstein: Style nit: I'm not a big fan of nested ternary operators. I'd prefer an if/else/else here.

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -133,19 +136,44 @@ export class ElementAnalyzer {
     }
   }
 
-  *eachAnalyzedAttribute(node: AnalyzableNode): Iterable<[string, string, AST.AttrNode | AST.HashPair]> {
+  private _throwIfInvalidAttribute(node: AnalyzableNode, property: AnalyzableProperty, name: string): void {
+    if (isPathExpression(property) && name === "class") {
+      const name = this._getAnalyzableAttributeName(property);
+      throw cssBlockError(`The ${name} attribute must contain a value and is not allowed to be purely positional. Did you mean ${name}="foo"?`, node, this.template);
```
chriseppstein: Can you add a test case that covers this condition?

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
abritinthebay: but... it's soooo ugly 😝 

Sure, no problem 😉

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -167,7 +202,9 @@ export class ElementAnalyzer {
 
     // Find the class or scope attribute and process it
     for (let [namespace, attrName, attribute] of this.eachAnalyzedAttribute(node)) {
-      if (attrName === "class") {
+      // should never be a path expression and "class" as that will throw an error before
+      // guard here is a) to get TS to agree b) just in case something, somehow, slips through.
+      if (attrName === "class" && !isPathExpression(attribute)) {
         this.processClass(namespace, attribute, element, forRewrite);
```
chriseppstein: A refactor of the code may invalidate the expectations of the comments here. If there's no good way to get the types to match our expectations, we should write this code as if we don't know about checking that is done elsewhere and do it again here...

```suggestion
      if (attrName === "class") {
         if (isPathExpression(attribute)) {
           throw this._missingClassAttributeValue(node, attribute, attrName);
         } else {
           this.processClass(namespace, attribute, element, forRewrite);
         }
```

But that assumes my assertion that "there's no good way to get the types to match our expectations" is true. Is it? I explored that question a bit in this follow-on PR: https://github.com/linkedin/css-blocks/pull/406

abritinthebay: Yeah, I have no idea if there is a way to get the types to agree here. In theory any `PathExpression` with class should have been removed in the previously called function, but I'm not sure how to convince typescript of that (it _can_ return a `PathExpression` but not one where the `attrName` is `class`)

It's almost like I need to create two types of PathExpression type just so I can mark it as returning the non-class-containing type to convince TS... but that seems a bit... weird.

abritinthebay: Is there some kind of Union Type magic that can work here? (if input is foo then output string bar will never be x?)

PR #404, opened 2020-02-26
feat(style-of): Errors if unsupported params have been passed.

### LINKED ISSUE #396: feat: style-of helper should error if a non-analyzed attribute or parameter is provided.
right now we just silently drop these arguments on the floor.

abritinthebay:
Fixes #396

Note: this builds off my PR #403 - for the positional support - so there are less changes in this PR than it looks. Any changes to that PR I'll port to this one before merging.

### DIFF
--- a/packages/@css-blocks/glimmer/src/Analyzer.ts
+++ b/packages/@css-blocks/glimmer/src/Analyzer.ts
@@ -62,15 +62,18 @@ export class GlimmerAnalyzer extends Analyzer<TEMPLATE_TYPE> {
   async analyze(dir: string, componentNames: string[]): Promise<GlimmerAnalyzer> {
     let components = new Set<string>(componentNames);
     this.debug(`Analyzing all templates starting with: ${componentNames}`);
-
-    for (let component of components) {
-      components.add(component);
-      let deps = this.resolver.recursiveDependenciesForTemplate(dir, component);
-      deps.forEach(c => components.add(c));
+    try {
+      for (let component of components) {
+        components.add(component);
+        let deps = this.resolver.recursiveDependenciesForTemplate(dir, component);
+        deps.forEach(c => components.add(c));
+      }
+    } catch (err) {
+      // if for some reason the resolver can't resolve. Otherwise it gets swallowed.
+      throw err;
     }
 
     this.debug(`Analyzing all components: ${[...components].join(", ")}`);
-
     for (let component of components) {
       await this.analyzeTemplate(dir, component);
     }
@@ -124,16 +127,18 @@ export class GlimmerAnalyzer extends Analyzer<TEMPLATE_TYPE> {
       MustacheStatement(node: AST.MustacheStatement) {
         if (!isAnalyzedHelper(node)) { return; }
         elementCount++;
+        const forbidNonBlockAttributes = isStyleOfHelper(node);
         const atRootElement = (elementCount === 1);
-        const element = elementAnalyzer.analyze(node, atRootElement);
+        const element = elementAnalyzer.analyze(node, atRootElement, forbidNonBlockAttributes);
         if (self.debug.enabled) self.debug(`MustacheStatement {{${node.path.original}}} analyzed:`, element.class.forOptimizer(self.cssBlocksOptions).toString());
       },
 
       SubExpression(node: AST.SubExpression) {
         if (!isStyleOfHelper(node)) { return; }
         elementCount++;
+        const forbidNonBlockAttributes = true;
         const atRootElement = (elementCount === 1);
-        const element = elementAnalyzer.analyze(node, atRootElement);
+        const element = elementAnalyzer.analyze(node, atRootElement, forbidNonBlockAttributes);
         if (self.debug.enabled) self.debug(`SubExpression (${node.path.original}) analyzed:`, element.class.forOptimizer(self.cssBlocksOptions).toString());
       },
 
--- a/packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
+++ b/packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
@@ -98,8 +98,8 @@ export class ElementAnalyzer {
     this.reservedClassNames = analysis.reservedClassNames();
   }
 
-  analyze(node: AnalyzableNode, atRootElement: boolean): AttrRewriteMap {
-    return this._analyze(node, atRootElement, false);
+  analyze(node: AnalyzableNode, atRootElement: boolean, forbidNonBlockAttributes = false): AttrRewriteMap {
+    return this._analyze(node, atRootElement, false, forbidNonBlockAttributes);
   }
 
   analyzeForRewrite(node: AnalyzableNode, atRootElement: boolean): AttrRewriteMap {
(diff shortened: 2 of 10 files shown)


packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -81,8 +81,8 @@ export class ElementAnalyzer {
     return this._analyze(node, atRootElement, false);
   }
 
-  analyzeForRewrite(node: AnalyzableNode, atRootElement: boolean): AttrRewriteMap {
-    return this._analyze(node, atRootElement, true);
+  analyzeForRewrite(node: AnalyzableNode, atRootElement: boolean, returnNonNamespaced = false): AttrRewriteMap {
```
chriseppstein: This check should also be performed during normal analysis of the style-of helper, not just at the point of rewriting (errors raised earlier are better since they'll save time before giving the user the feedback).

packages/@css-blocks/glimmer/src/Rewriter.ts
```diff
@@ -250,10 +250,11 @@ export class GlimmerRewriter implements ASTPluginWithDeps {
   StyleOfHelper(node: AST.MustacheStatement | AST.SubExpression): void {
     this.elementCount++;
     let atRootElement = (this.elementCount === 1);
-    let attrMap = this.elementAnalyzer.analyzeForRewrite(node, atRootElement);
+    let attrMap = this.elementAnalyzer.analyzeForRewrite(node, atRootElement, true);
     let attrNames = Object.keys(attrMap);
     if (attrNames.length !== 1 || attrNames[0] !== "class") {
-      console.error("Internal Error: unexpected attributes in rewrite for style-of helper", attrNames);
+      const names = attrNames.filter(name => name !== "class");
+      throw cssBlockError(`Unexpected attribute(s) [${names.join(", ")}] in rewrite for style-of helper.`, node, this.template);
```
chriseppstein: The attribute rewrite mapping is being over-used to serve two purposes here. In the general case, these two purposes could collide and want to set values that conflict with each other. At the very least, to return this information I'd rather the return type become a meta object that returns invalid attribute nodes as one property and the attribute rewrite as another property.

That said, I think we don't even need to return those invalid attributes to the style-of helper itself. Instead we should take a parameter like `forbidNonBlockAttributes` which would default to false. Then when we encounter a non block attribute and that parameter is true, we raise an error on the specific illegal attribute node (rather than underlining the entire helper as using the node in this context would do).

packages/@css-blocks/glimmer/src/utils.ts
```diff
@@ -5,6 +5,13 @@ import { ClassifiedParsedSelectors } from "opticss";
 
 import { TEMPLATE_TYPE } from "./Template";
 
+// GlimmerTemplates.ResolvedFile can have path, Opticss.Template doesn't.
+// template.path, if it exists, gives the exact path of the template.
+// template.identifier is less specific.
+interface Template extends TemplateInfo<TEMPLATE_TYPE> {
+  path?: string;
+}
+
```
chriseppstein: I think we don't need to do this or the signature change below if we take the approach I discussed above.

abritinthebay: In general I've found it WAY more useful to use `path` than `identifier`. Identifier is functionally useless to me in a an error message, worse: it references column and row numbers of _something it's not actually referring to_ if it uses `identifier`. 

I made this change to help track down a problem that `identifier` was not useful for. 

If we don't want to use path we should remove the column and row number at the very least. It's just giving wrong information.

packages/@css-blocks/glimmer/src/Rewriter.ts
abritinthebay: Ah got it. Yeah, I originally went that sort of route but specifically wanted to return _all_ the attributes that were wrong (avoiding wack-a-mole debugging).

I see your point tho - it's somewhat overloading this.

packages/@css-blocks/glimmer/src/utils.ts
chriseppstein: Ah. I thought this existed for a different reason. The concrete implementation of TemplateInfo always has a `path` set. I think we can just put `path?: string | undefined | null` on the `TemplateInfo` interface.

packages/@css-blocks/glimmer/src/Rewriter.ts
chriseppstein: In general, we need to make template parsing fault tolerant and return multiple errors instead of just one. We can address that in a separate PR.

packages/@css-blocks/glimmer/src/Analyzer.ts
```diff
@@ -62,15 +62,18 @@ export class GlimmerAnalyzer extends Analyzer<TEMPLATE_TYPE> {
...
-      deps.forEach(c => components.add(c));
+    try {
+      for (let component of components) {
+        components.add(component);
+        let deps = this.resolver.recursiveDependenciesForTemplate(dir, component);
+        deps.forEach(c => components.add(c));
+      }
+    } catch (err) {
+      // if for some reason the resolver can't resolve. Otherwise it gets swallowed.
+      throw err;
```
chriseppstein: I'm skeptical that this try/catch block is necessary. Any exception raised within the async function should reject the promise... catching and rethrowing it doesn't change anything. Can you add a test case for the issue you're trying to solve with this code?

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
```diff
@@ -210,6 +210,11 @@ export class ElementAnalyzer {
               } else {
                 yield { type: "state", namespace, name: attrName, property };
               }
+            } else {
+              if (forbidNonBlockAttributes) {
+                // We shouldn't have any properties that aren't namespaced!
+                throw cssBlockError(`Unexpected un-namespaced attribute "${name}".`, node, this.template);
```
chriseppstein: Can we change this message to: `An attribute without a block namespace is forbidden in this context: ${name}`

packages/@css-blocks/glimmer/src/Analyzer.ts
abritinthebay: Will test. It certainly helped why I was testing but I wonder if there was a knock-on effect happening with something else. Will see.

packages/@css-blocks/glimmer/src/ElementAnalyzer.ts
abritinthebay: Sure 👍

packages/@css-blocks/glimmer/src/Analyzer.ts
abritinthebay: Can confirm - without it it swallows the error. I only noticed this when I accidentally forced it to load a file that didn't exist. Seems like an edge case but... it seems to silently fail otherwise.

chriseppstein: I took this change out. It shouldn't be necessary. The tests still pass. I wish I could reproduce the issue you saw...

PR #405, opened 2020-02-26
chore(deps): Updates a bunch of dependancies.

abritinthebay:
collates together #390 #388 #379 #362 #361 and #360    

Updates the deps that aren't in the ember-cli project (because there are changes incoming there), playground, or website. These have been tested and seem to work fine.


chriseppstein:
This is failing in node 8 :(

PR #418, opened 2020-03-10
Drop node 6, node 8 and node 11 support.

chriseppstein:
This PR officially removes our node 8 support and updates our typescript configuration to target output that takes advantage of node 10+ language features.

It also cleans up a few other weird things I noticed:
* we had the wrong license in some of our packages.
* sorts out some build failures that I got when I downgraded `@types/node` from node 11 to node 10. 
* cleans up a weird build issue in the website (had to downgrade eslint).


packages/@css-blocks/jsx/tsconfig-runtime.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     // Compilation Configuration
```
abritinthebay: comments are technically invalid json, so... probably should remove this at some point. Not a blocker because it's pre-existing but... a thought.

chriseppstein: Yeah, tsconfig files [support comments](https://github.com/microsoft/TypeScript/pull/5450).

That said, this comment isn't adding any value so I'm not opposed to removing it.

PR #432, opened 2020-04-03
Optional Preprocessors & library/application API contract.

chriseppstein:
There is now a core data type for preprocessors that conditionally process a stylesheet. The motivation for this is to allow libraries to expose a preprocessor integration for their specific Sass + Eyeglass library.

This allows the library to maintain some semblance of control over how the compilation is performed, in case that is important to that module (E.g. the precision value for Sass or exposing assets via eyeglass).

There's a base class for libraries to use (`DirectoryScopedPreprocessor`), and a helper function for applications that integrates all of the preprocessor adaptors into a single processor that supplies sass, eyeglass, and application default options.

If the `DirectoryScopedPreprocessor` isn't sufficient for a library's needs, they can implement the `PreprocessorProvider` interface.

Applications can use the simpler, `OptionalAdaptor` function for creating their own optional adaptors for preprocessing their own block files in different ways.

The README for `@css-blocks/eyeglass` is updated to include boilerplate code that libraries and applications can use to get started.

### DIFF
--- a/packages/@css-blocks/core/src/BlockTree/RulesetContainer.ts
+++ b/packages/@css-blocks/core/src/BlockTree/RulesetContainer.ts
@@ -100,6 +100,8 @@ export class RulesetContainer<S extends Styles> {
     let style = this.parent;
     let selectors: ParsedSelector[] = style.getParsedSelectors(rule);
 
+    // XXX I think this is wrong. if the selectors target different styles it
+    // will get confused. Need to add tests for that.
     selectors.forEach((selector) => {
       let ruleSet = new Ruleset(file, rule, style);
       let key = selector.key;
--- a/packages/@css-blocks/eyeglass/src/index.ts
+++ b/packages/@css-blocks/eyeglass/src/index.ts
@@ -1,18 +1,30 @@
-import { ProcessedFile } from "@css-blocks/core";
-import Eyeglass from "eyeglass"; // works, even tho a cjs export. huh.
-import { Options, Result, SassError } from "node-sass";
-import SassImplementation from "node-sass";
+import type { OptionalPreprocessor, Preprocessor, ProcessedFile, ResolvedConfiguration } from "@css-blocks/core";
+import type { EyeglassOptions, default as Eyeglass } from "eyeglass"; // works, even tho a cjs export. huh.
+import type { Result, SassError } from "node-sass";
+import type SassImplementation from "node-sass";
+import { sep as PATH_SEPARATOR } from "path";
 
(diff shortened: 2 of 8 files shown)


packages/@css-blocks/core/src/BlockTree/RulesetContainer.ts
```diff
@@ -100,6 +100,8 @@ export class RulesetContainer<S extends Styles> {
     let style = this.parent;
     let selectors: ParsedSelector[] = style.getParsedSelectors(rule);
 
+    // XXX I think this is wrong. if the selectors target different styles it
+    // will get confused. Need to add tests for that.
```
abritinthebay, line 104: Do we have an issue for this?

packages/@css-blocks/eyeglass/src/index.ts
```diff
@@ -1,18 +1,30 @@
-import { ProcessedFile } from "@css-blocks/core";
-import Eyeglass from "eyeglass"; // works, even tho a cjs export. huh.
-import { Options, Result, SassError } from "node-sass";
-import SassImplementation from "node-sass";
+import type { OptionalPreprocessor, Preprocessor, ProcessedFile, ResolvedConfiguration } from "@css-blocks/core";
+import type { EyeglassOptions, default as Eyeglass } from "eyeglass"; // works, even tho a cjs export. huh.
```
abritinthebay, line 2: I think that's by design as its a type - and so TS looks for types...? I remember reading something about that.

chriseppstein: `import type` is new in 3.8. It ensures that you don't end up importing the concrete values from the library, you get an error if you try to import any value type. For things like node-sass where we've got it installed for development but don't want to have any runtime dependency, the `import type` statement gives me a little peace of mind.

packages/@css-blocks/core/src/BlockTree/RulesetContainer.ts
chriseppstein: I don't think so.

packages/@css-blocks/eyeglass/src/index.ts
chriseppstein: Ah, I think you're commenting on the comment about cjs... that was a comment [you wrote](https://github.com/linkedin/css-blocks/commit/6f92d19c4362ca5e0b3971977d645eda5682c928#diff-b4b0634b9f9d4a9bee536ebda66c3521R2) in your initial implementation :P

PR #453, opened 2020-06-09
feat: Parse and set block-interface-index

zimmi88:
- When processing definition files, recognize and set the interface
index on style nodes in the Block. As with block-class, if each style
node found has not declared an interface-index by the end of processing,
it's an error.
- Test coverage for ensuring block-interface-index is declared.

### DIFF
--- a/packages/@css-blocks/core/src/BlockParser/features/add-interface-indexes.ts
+++ b/packages/@css-blocks/core/src/BlockParser/features/add-interface-indexes.ts
@@ -8,11 +8,11 @@ import { getStyleTargets } from "../block-intermediates";
 import { stripQuotes } from "../utils";
 
 /**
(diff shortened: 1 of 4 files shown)


packages/@css-blocks/core/src/BlockParser/features/add-interface-indexes.ts
```diff
@@ -0,0 +1,88 @@
...
+            sourceRange(configuration, root, file, decl),
+          ),
+        );
+      } else {
+        foundIndexes.push(parsedIndex);
+      }
+
+      // Set the index on the related style node.
+      const parsedSelectors = block.getParsedSelectors(rule);
+      parsedSelectors.forEach(sel => {
```
chriseppstein, line 62: It's an error if there's more than one selector with rule containing interface-index.

PR #454, opened 2020-06-16
Block definition compiler

chriseppstein:
This PR gets the basics of block definition compilation working. Two aspects of block definition files are omitted from this PR (`inherited-styles` and conflict resolutions). These aspects are important for pre-compilation but don't block our current project.

In this PR I introduce a CSS-Blocks specific AST that describes both legal CSS Block files as well as legal Block Definition Files. I use this AST first to construct a partial CSS Block AST during block parsing this is then "mapped" to a definition AST during block compilation. The rest of the AST is built up by examining the Block data model. Once the block definition AST is constructed I use an AST visitor to create the postcss AST which is then stringified.

### DIFF
--- a/packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/PostcssASTBuilder.ts
+++ b/packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/PostcssASTBuilder.ts
@@ -1,0 +1,103 @@
+import { postcss } from "opticss";
+
+import { BlockExport, BlockReference, BlockSyntaxVersion, DefinitionAST, GlobalDeclaration, LocalBlockExport, Name, Rename, Rule, Visitor, visit } from "../../BlockParser/ast";
+import { BLOCK_GLOBAL } from "../../BlockSyntax";
+
+import { SelectorASTBuilder } from "./SelectorASTBuilder";
+
+export class PostcssASTBuilder implements Visitor<DefinitionAST> {
+  postcss: typeof postcss;
+  root: postcss.Root;
+  currentRule: postcss.Rule | undefined;
+  constructor(postcssImpl: typeof postcss) {
+    this.postcss = postcssImpl;
+    this.root = this.postcss.root();
+  }
+
+  namedReferences(references: Array<Name | Rename>): string {
+    let result = "";
+    result += "(";
+    let prevRef = false;
+    for (let ref of references) {
+      if (prevRef) {
+        result += ", ";
+      }
+      result += ref.name;
+      if (ref.asName) {
+        result += ` as ${ref.asName}`;
+      }
+      prevRef = true;
+    }
+    result += ")";
+    return result;
+  }
+
+  BlockSyntaxVersion(blockSyntaxVersion: BlockSyntaxVersion): void {
+    this.root.append(this.postcss.atRule({
+      name: "block-syntax-version",
+      params: blockSyntaxVersion.version.toString(),
+    }));
+  }
+
+  BlockReference(blockReference: BlockReference): void {
+    let params = "";
+    if (blockReference.defaultName) {
+      params += blockReference.defaultName;
+    }
+    if (blockReference.defaultName && blockReference.references) {
+      params += ", ";
+    }
+    if (blockReference.references) {
+      params += this.namedReferences(blockReference.references);
+    }
+    params += ` from "${blockReference.fromPath}"`;
+    let atRule = this.postcss.atRule({
+      name: "block",
+      params,
+    });
+    this.root.append(atRule);
+  }
+
+  LocalBlockExport(localBlockExport: LocalBlockExport): void {
+    let params = this.namedReferences(localBlockExport.exports);
+    let atRule = this.postcss.atRule({
+      name: "export",
+      params,
+    });
+    this.root.append(atRule);
+  }
+
+  BlockExport(blockExport: BlockExport): void {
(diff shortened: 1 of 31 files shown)


packages/@css-blocks/core/src/BlockCompiler/index.ts
```diff
@@ -15,20 +15,76 @@ import {
...
     this.postcss = postcssImpl;
   }
 
+  setDefinitionCompiler(definitionCompiler: BlockDefinitionCompiler) {
+    this.definitionCompiler = definitionCompiler;
+  }
+
+  compileWithDefinition(block: Block, root: postcss.Root, reservedClassNames: Set<string>, definitionPath: typeof INLINE_DEFINITION_FILE): CompiledBlockAndInlineDefinition;
+  compileWithDefinition(block: Block, root: postcss.Root, reservedClassNames: Set<string>, definitionPath: string): CompiledBlockAndDefinition;
+  compileWithDefinition(block: Block, root: postcss.Root, reservedClassNames: Set<string>, definitionPath: string | typeof INLINE_DEFINITION_FILE): CompiledBlockAndDefinition | CompiledBlockAndInlineDefinition {
```
zimmi88, line 54: Given this is the surface-level compiler API that most things outside of core would interact with, can we add a comment header to this method?

packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/PostcssASTBuilder.ts
```diff
@@ -0,0 +1,103 @@
...
+    let params = this.namedReferences(blockExport.exports);
+    params += ` from "${blockExport.fromPath}"`;
+    let atRule = this.postcss.atRule({
+      name: "export",
+      params,
+    });
+    this.root.append(atRule);
+  }
+
+  Rule(rule: Rule<DefinitionAST>): void | boolean {
```
zimmi88, line 80: This always returns `false`.

packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      let declarations = new Array<Declaration>();
+      declarations.push(builders.declaration("composes", composition.paths.join(", ")));
+      rules.push(builders.rule([selector], declarations));
+    }
+    return rules;
+  }
+
+  blockReferences(root: postcss.Root, block: Block): void {
+    block.eachBlockReference((name, _block) => {
+      root.append(postcss.atRule({name: "block", params: `${name} from ""`}));
```
zimmi88, line 139: No path?

packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    });
+  }
+
+  insertReference(css: postcss.Root, definitionPath: string) {
```
zimmi88, line 143: This might be confused with block references. Maybe `insertDefinitionUrlComment()`?

(same with `insertInlineReference` to... `insertEmbeddedDefinitionDataComment()`?)

packages/@css-blocks/core/src/BlockTree/Block.ts
```diff
@@ -38,6 +38,8 @@ import { Styles } from "./Styles";
 export class Block
   extends Inheritable<Block, Block, never, BlockClass> {
 
+  public blockAST?: BlockAST;
+  public definitionAST?: DefinitionAST;
```
zimmi88, line 42: When do these get instantiated? Do these need to be established when reconstituting a block from a definition file, or only when compiling?

packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/PostcssASTBuilder.ts
chriseppstein: That's intentional. Returning false prevents descent of the visitor into the child nodes. In this case there's no need to visit those nodes because they are traversed explicitly in this method.

packages/@css-blocks/core/src/BlockTree/Block.ts
chriseppstein: `definitionAST` doesn't get used right now so I deleted it. `blockAST` gets set during block parse time.

In an ideal world, now that we have an AST, we'd fully parse to the AST and then use that AST to build the Block data model. But we don't yet live in that world. If we did though, we'd set the appropriate AST during parse based on which type of block was parsed.

packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/index.ts
chriseppstein: Good idea. Done.

packages/@css-blocks/core/src/BlockCompiler/BlockDefinitionCompiler/index.ts
chriseppstein: whoops!

packages/@css-blocks/core/src/BlockCompiler/index.ts
chriseppstein: Done. I also added an API comment for the `compile` method.

PR #455, opened 2020-06-21
Definition File Full Workflow - Acceptance Tests

zimmi88:
This PR includes two commits...

**fix: Allow quotes surrounding block-name.**
    
Definition file output wraps the block-name in quotes, but there are
existing rules in CSS Blocks disallowing this. This removes this rule,
using stripQuotes() to unwrap the block-name value. With this change,
block-name, block-id, and block-class will all behave in the same way.

**test: Definition Files Basic Acceptance Tests.**
    
- Test the full workflow of working with definition files. From a source
block file, compile it into css and definition data, then reconstitute
the block and validate the name, GUID, and properties match.
- Tests for both external definition files and embedded definition data.
- Update the mock importer to recognize embedded definition data.
- Add some comments to MockImporter and the registry.

------------------------

This PR only encompasses basic acceptance coverage using a source CSS Blocks file without any imports, exports, or aliases.

CI is going to fail because of missing types for `ember-cli-htmlbars`. Manually verified that `yarn test` in the `/core` directory is passing locally.

### DIFF
--- a/packages/@css-blocks/core/src/BlockCompiler/index.ts
+++ b/packages/@css-blocks/core/src/BlockCompiler/index.ts
@@ -18,7 +18,7 @@ import {
 import { BlockDefinitionCompiler, INLINE_DEFINITION_FILE } from "./BlockDefinitionCompiler";
 import { ConflictResolver } from "./ConflictResolver";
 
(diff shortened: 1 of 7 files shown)


zimmi88:
^ rebased off of latest `ember-rewrite`

PR #462, opened 2020-07-24
Ember v2 caching and deserialization.

chriseppstein:
This PR implements the following:

* caching for `@css-blocks/ember` so that when templates are cached, the corresponding css blocks output files are also cached.
* Extract utilities that are needed to be shared between `@css-blocks/ember` and `@css-blocks/ember-app` to a npm package named `@css-blocks/ember-support`.
* Deserialize template analysis and precompiled css block files.
* Optimize css and emit optimized css to an output file.
* Basic scaffolding and build structure for the ember runtime data, service and helpers.


packages/@css-blocks/ember-app/runtime/app/services/css-blocks.ts
```diff
@@ -0,0 +1,14 @@
...
+import Service from "@ember/service";
+
+/// @ts-ignore
+import { data } from "./-css-blocks-data";
+
+// tslint:disable-next-line:no-default-export
+export default class CSSBlocksService extends Service {
+  classNamesFor(_args: Array<string | number | boolean | null>): string {
+    return data.className;
+  }
```
timlindvall, line 13: placeholder?

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
```diff
@@ -0,0 +1,126 @@
...
+  constructor(appName: string, inputNodes: InputNode[], cssBlocksOptions: CSSBlocksOptions, options?: PluginOptions) {
+    super(mergeTrees(inputNodes), options || {});
+    this.appName = appName;
+    this.previousSourceTree = new FSTree();
+    this.cssBlocksOptions = cssBlocksOptions;
+  }
+  processString(contents: string, _relativePath: string): string {
+    return contents;
+  }
+  async build() {
```
timlindvall, line 45: Huge function that could use some commentary.

packages/@css-blocks/ember-support/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@css-blocks/ember-support",
```
timlindvall, line 2: New module?

packages/@css-blocks/ember/runtime/app/helpers/-css-blocks.ts
```diff
@@ -0,0 +1,18 @@
+/// @ts-ignore
```
timlindvall, line 1: Why does `ember` need a runtime? Wouldn't the runtime be `ember-app`'s responsibility?

packages/@css-blocks/ember/src/CSSBlocksTemplateCompilerPlugin.ts
```diff
@@ -0,0 +1,350 @@
...
+ *
+ * Note: It is possible for several templates to depend on the same CSS Block
+ * file and so their caches will point to the same additional file cache
+ * entries. This produces a little extra work when the cache is warm but
+ * ensures consistency if one of the templates is removed.
+ *
+ * In the case where just one of the templates is invalidated, the css block file
+ * will get recompiled after it is retrieved from cache, but this is ok because
+ * the output from css-blocks will be identical.
+ */
```
timlindvall, line 71: Thanks for the summary on this! 👍

private-packages/fixtures-ember-v2/ember-app.code-workspace
```diff
@@ -23,9 +27,13 @@
 		{
 			"path": "ember-lazy-engine",
 			"name": "Fixture: Ember Lazy Engine"
+		},
+		{
+			"path": "npm-based-blocks",
+      "name": "Fixture: npm-based-blocks"
```
timlindvall, line 33: nitpick: ------> indent

packages/@css-blocks/core/src/Analyzer/Analysis.ts
```diff
@@ -161,19 +172,35 @@ export class Analysis<K extends keyof TemplateTypes> {
...
-    for (let name of names) {
-      let superBlock = this.blocks[name].base;
-      while (superBlock) {
-        if (superBlock === block) return name === DEFAULT_EXPORT ? "" : name;
-        superBlock = superBlock.base;
+    return null;
+  }
+
+  _searchForBlock(blockToFind: Block, block: Block, parentPath: string): string | null {
+    if (block === blockToFind || block.isAncestorOf(blockToFind)) {
```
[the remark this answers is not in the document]
chriseppstein: Yeah, I don't remember why I added that and it doesn't seem necessary. Good catch.

packages/@css-blocks/core/src/Analyzer/Analysis.ts
```diff
@@ -161,19 +172,35 @@ export class Analysis<K extends keyof TemplateTypes> {
...
+    let blockRefs = new Array<[string, Block]>();
+    block.eachBlockReference((name, refBlock) => {
+      blockRefs.push([name, refBlock]);
+    });
+
+    for (let [name, refBlock] of blockRefs) {
+      let currentSearchPath = `${parentPath}>${name}`;
+      let rv = this._searchForBlock(blockToFind, refBlock, currentSearchPath);
+      if (rv !== null) {
+        return rv;
```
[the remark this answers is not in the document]
chriseppstein: It's an acyclic directed graph. We don't currently maintain any references of the inverse relationships. E.g. (`referencedBy`, `exportedBy`, `inheritedBy`, `implementedBy`). I'm generally averse to creating data models that have circular references because it makes serialization and deserialization much more complicated and increases the likelihood of inadvertently coding an infinite loop.

packages/@css-blocks/core/src/Analyzer/Analysis.ts
```diff
@@ -329,14 +385,122 @@ export class Analysis<K extends keyof TemplateTypes> {
...
+  }
 
-    // Serialize all discovered Elements.
-    this.elements.forEach((el, key) => {
-      elements[key] = el.serialize(styleIndexes);
+  /**
+   * Creates a TemplateAnalysis from its serialized form.
+   * @param serializedAnalysis The analysis to be recreated.
+   * @param options The plugin options that are used to parse the blocks.
+   * @param postcssImpl The instance of postcss that should be used to parse the block's css.
```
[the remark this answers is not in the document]
chriseppstein: Fixed for both methods.

packages/@css-blocks/core/src/Analyzer/Analysis.ts
```diff
@@ -329,14 +385,122 @@ export class Analysis<K extends keyof TemplateTypes> {
...
+        } else {
+          assertNever(analyzedStyle);
+        }
+      }
+      element.seal();
+      analysis.elements.set(elID, element);
+    });
+
+    // tslint:disable-next-line:prefer-unknown-to-any
+    return analysis;
```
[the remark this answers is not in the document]
chriseppstein: Nope. The comment was unnecessary. I removed it.

packages/@css-blocks/core/src/Analyzer/Analysis.ts
```diff
@@ -329,14 +385,122 @@ export class Analysis<K extends keyof TemplateTypes> {
...
-    // Serialize all discovered Elements.
-    this.elements.forEach((el, key) => {
-      elements[key] = el.serialize(styleIndexes);
+  /**
+   * Creates a TemplateAnalysis from its serialized form.
+   * @param serializedAnalysis The analysis to be recreated.
+   * @param options The plugin options that are used to parse the blocks.
+   * @param postcssImpl The instance of postcss that should be used to parse the block's css.
+   */
+  static async deserializeSource (
```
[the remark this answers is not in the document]
chriseppstein: Thanks for the nudge. I refactored it such that deserialization of each element analysis occurs in `ElementAnalysis` next to the serialization code that is there. I think this makes the code a big more digestible and the proximity to the serialization code clears up some of the confusion as well.

I also added more comments.

packages/@css-blocks/core/src/Analyzer/ElementAnalysis.ts
```diff
@@ -41,8 +41,15 @@ export interface HasAttrValue<T extends Style = Style> {
   value: Set<T>;
 }
 
-export function isBooleanAttr(o: object): o is HasAttrValue {
-  return !!(<HasAttrValue>o).value;
+export function isBooleanAttr(o: object): o is HasAttrValue | SerializedHasAttrValue {
+  return hasAttrValue(o);
+}
```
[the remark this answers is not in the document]
chriseppstein: I realized while working on this PR, that `isBooleanAttr` was a misnomer for what it was really doing so I renamed it to `hasAttrValue` but then I added `isBooleanAttr` back so that code wouldn't break.

I checked all the existing uses of `isBooleanAttr` and they all worked better by calling them `hasAttrValue` so now I've completely removed `isBooleanAttr`.

packages/@css-blocks/core/src/BlockParser/BlockFactory.ts
```diff
@@ -362,9 +362,10 @@ export class BlockFactory {
     }
 
     // Construct a Block out of the definition file.
-    const block = await this.parser.parseDefinitionSource(definitionAst, file.definitionIdentifier, file.blockId, file.defaultName);
+    const block = await this.parser.parseDefinitionSource(definitionAst, file.identifier, file.blockId, file.defaultName);
```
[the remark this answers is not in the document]
chriseppstein: This code was resulting in the canonical identifier for the block being the definition file's identifier. But that's incorrect. The identifier of the compiled css is the canonical identifier of the file, when it was using the definition file's identifier i ended up with lots of bugs because the identifier returned wasn't the one that was requested of the `BlockFactory`.

packages/@css-blocks/ember-app/runtime/app/services/css-blocks.ts
chriseppstein: yes. just to show the dots are getting connected. I will replace it with a real implementation in the next PR.

packages/@css-blocks/ember/runtime/app/helpers/-css-blocks.ts
chriseppstein: The `ember` addon does the rewrite which invokes this helper; it felt proper to keep those bundled together so that if the API to the helper needs to change, it changes the call sites at the same time.

The actual rewriting occurs in a service that is bundled in the `ember-app` addon, which uses the data produced by that addon so keeping those bundled together also feels like the right thing to do.

packages/@css-blocks/ember/src/CSSBlocksTemplateCompilerPlugin.ts
chriseppstein: 😃

private-packages/fixtures-ember-v2/ember-app.code-workspace
chriseppstein: tabs vs spaces issue 😬

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
chriseppstein: I will refactor this in a followup PR.

PR #492, opened 2020-08-29
Update concat strategy; Fix sourcemap support in Ember v2 pipeline

### LINKED ISSUE #476: Compiled Blocks from CSS from addons ends up in vendor.css
Those files should be pruned out of the build because those styles get added to the application's output.

### LINKED ISSUE #485: Sourcemap Support
Sourcemap support got dropped during the rewrite, we need to track down what's wrong and get it working correctly again.

timlindvall:
Fixes #476. Fixes #485.


packages/@css-blocks/ember/src/index.ts
```diff
@@ -94,7 +100,7 @@ const EMBER_ADDON: AddonImplementation<CSSBlocksAddon> = {
     // We compile CSS Block files in the template tree, so in the CSS Tree all
     // we need to do is prune them out of the build before the tree gets
     // built.
-    return withoutCssBlockFiles(tree);
+    return withoutCssBlockFiles(tree, true);
```
chriseppstein, line 103: I noticed this the other day. Thank you for fixing it!

ramitha (approved):
Does this mean we won't need the the `import 'css-blocks'` within the app.(s)css file?

timlindvall:
> Does this mean we won't need the the `import 'css-blocks'` within the app.(s)css file?

Yes, `broccoli-concat` will merge the files together. We'll provide the ability to override concat settings in the event there's multiple files to merge together or apps want to change the order.

packages/@css-blocks/ember-app/src/index.ts
```diff
@@ -182,14 +187,66 @@ const EMBER_ADDON: AddonImplementation<CSSBlocksApplicationAddon> = {
...
 };
 
+/**
+ * Build a default broccoli-concat config, using given enviroment settings.
+ * @param env - The addon environment.
+ * @returns - Default broccoli-concat options, accounting for current env settings.
+ */
+function buildDefaultBroccoliConcatOptions(env: AddonEnvironment): BroccoliConcatOptions {
+  return {
+    inputFiles: ["assets/css-blocks.css", `assets/${env.modulePrefix}.css`],
```
chriseppstein, line 239: This should use the output of `cssBlocksOutputFilename()` instead of hardcoding `assets/css-blocks.css` so that it respects the `output` filename configuration.

packages/@css-blocks/ember-app/src/index.ts
```diff
@@ -182,14 +187,66 @@ const EMBER_ADDON: AddonImplementation<CSSBlocksApplicationAddon> = {
...
+      let concatOptions: BroccoliConcatOptions;
+      if (env.config.broccoliConcat) {
+        concatOptions = Object.assign(
+          {},
+          buildDefaultBroccoliConcatOptions(env),
+          env.config.broccoliConcat,
+        );
+      } else {
+        concatOptions = buildDefaultBroccoliConcatOptions(env);
+      }
```
chriseppstein, line 215: It's slightly less efficient, but I think we should move the conditional check into the expression, it cleans up some of the duplication and makes this block easier to read.

```suggestion
      let concatOptions: BroccoliConcatOptions = Object.assign(
        {},
        buildDefaultBroccoliConcatOptions(env),
        env.config.broccoliConcat || {},
      );
```

packages/@css-blocks/ember-utils/src/options.ts
```diff
@@ -19,12 +70,14 @@ export interface ResolvedCSSBlocksEmberOptions {
   analysisOpts: AnalysisOptions;
   parserOpts: ParserOptions;
   optimization: Partial<OptiCSSOptions>;
+  broccoliConcat: BroccoliConcatOptions | "SKIP";
```
chriseppstein, line 73: I prefer `false` to the string literal `"SKIP"` here. It means that down on line 80 you'll need to check for `undefined` explicitly, but I think it's a more standard API for the user.

Alternatively, you can add an option like `disableBroccoliConcat` that is at the same level as `broccoliConcat` and it would cause the concat to be skipped if set to true, even if concat options were provided.

packages/@css-blocks/ember-utils/src/options.ts
```diff
@@ -5,12 +5,63 @@ import type { OptiCSSOptions } from "opticss";
...
+   * The files to contatenate together during the postprocess step. Do keep
+   * in mind this you'll have access only to the files in the CSS postprocess
+   * tree for ember-app. You can use globs if preferred.
+   *
+   * Any top-level files that are processed by Ember CLI will likely be stored
+   * at "styles/<name-of-file>.css". You can use a plugin such as broccoli-debug
+   * to determine the structure of your styles after processing.
+   *
+   * Your CSS Blocks file will be available by default
+   * at "styles/css-blocks.css", but will vary if the "output" property is set
```
chriseppstein, line 41: I thought these files were in the `assets` directory instead of the `styles` directory... that's what we're using in the config defaults at least.

```suggestion
   * at "assets/<name-of-file>.css". You can use a plugin such as broccoli-debug
   * to determine the structure of your styles after processing.
   *
   * Your CSS Blocks file will be available by default
   * at "assets/css-blocks.css", but will vary if the "output" property is set
```

chriseppstein (changes requested):
This is great stuff. I'm glad you got this working, it's super important to the developer experience. I think it needs a few small changes and then it will be ready to merge.

packages/@css-blocks/ember-app/src/index.ts
```diff
@@ -182,14 +187,66 @@ const EMBER_ADDON: AddonImplementation<CSSBlocksApplicationAddon> = {
...
+        return tree;
+      }
+
+      // Merge default concat options with the options provided by the app.
+      let concatOptions: BroccoliConcatOptions;
+      if (env.config.broccoliConcat) {
+        concatOptions = Object.assign(
+          {},
+          buildDefaultBroccoliConcatOptions(env),
+          env.config.broccoliConcat,
```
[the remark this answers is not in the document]
timlindvall: They may want to disable sourcemaps altogether, or they may want to make the sourcemap inline, or some other config I'm not thinking of. I think my goal here was to muck with what the user gives as minimally as possible, but it's true things would break if we omit some things.

I'll update this to be smarter about how we're merging this object together.

packages/@css-blocks/ember-app/src/index.ts
```diff
@@ -224,19 +217,55 @@ const EMBER_ADDON: AddonImplementation<CSSBlocksApplicationAddon> = {
...
+  );
+}
+
 /**
  * Build a default broccoli-concat config, using given enviroment settings.
  * @param env - The addon environment.
  * @returns - Default broccoli-concat options, accounting for current env settings.
  */
 function buildDefaultBroccoliConcatOptions(env: AddonEnvironment): BroccoliConcatOptions {
+  const cssBlocksOutputFilename = env.config.output || "css-blocks.css";
```
chriseppstein, line 262: I'd prefer that we didn't duplicate the logic for accessing this option with a default. Is there a reason why the function I suggested couldn't be used?

chriseppstein (approved):
One comment. Approved to merge once that's addressed.

packages/@css-blocks/ember-app/src/index.ts
timlindvall: The paths are different. The aforementioned method presumes `app/styles/`, but the path should be `assets/`.

PR #498, opened 2020-09-10
Detect generated class name conflicts

### LINKED ISSUE #496: Detect Generated Classname Conflicts
We need to raise an error if a generated classname from opticss conflicts with a css class used in a stylesheet that isn't from css-blocks.

We should scan all css files that aren't from us to make sure their classnames aren't used by the optimizer. (ideally we'd register these with the optimizer so that it avoids it but it seems like it's too late to do that in the ember-cli build process).

This should only run during production builds, when opticss is enabled. we can skip this whenever opticss is disabled.

timlindvall:
Resolves #496.

### Changes
 - Scan for class names used in application CSS not otherwise parsed by CSS Blocks.
- Throw an error if there's a conflict between the optimizer's class names and application CSS.
- Add a new config option `appClasses` that's an alias for the optimizer config option that omits class names.
- Add a smoke test to the ember-app fixture to exercise this new code.
- Some refactoring to move common methods, interfaces, constants to shared utility files.

### Testing Done
All testing done in the v2 ember-app fixture.
- Build fails if optimization is enabled and the `.i` class is present.
- Build succeeds if optimization is enabled, the `.i` class is present, and `appClasses` is declared.

### Sample Error Message
```
Build Error (CSSBlocksStylesPostprocessorPlugin)

Your application CSS contains classes that are also generated by the CSS optimizer. This can cause style conflicts between your application's classes and those generated by CSS Blocks.
To resolve this conflict, you should add any short class names in non-block CSS (~5 characters or less) to the list of disallowed classes in your build configuration.
(You can do this by setting css-blocks.appClasses to an array of disallowed classes in ember-cli-build.js.)

Conflicting classes:
i
```


packages/@css-blocks/ember-app/src/broccoli-plugin.ts
```diff
@@ -183,10 +213,139 @@ export class CSSBlocksStylesPreprocessorPlugin extends Plugin {
...
+    this.output.writeFileSync("assets/app-classes.log", logfile);
+    debug("Wrote log file to broccoli tree.");
+
+    if (collisions.length > 0) {
+      throw new Error(
+        "Your application CSS contains classes that are also generated by the CSS optimizer. This can cause style conflicts between your application's classes and those generated by CSS Blocks.\n" +
+        "To resolve this conflict, you should add any short class names in non-block CSS (~5 characters or less) to the list of disallowed classes in your build configuration.\n" +
+        "(You can do this by setting css-blocks.appClasses to an array of disallowed classes in ember-cli-build.js.)\n\n" +
+        "Conflicting classes:\n" +
+        collisions.reduce((prev, curr) => prev += `${curr}\n`, ""),
```
chriseppstein, line 346: It would be nice if this told the user where the conflicting classname was found so they could decide which to change.

packages/@css-blocks/ember-utils/src/options.ts
```diff
@@ -49,19 +49,42 @@ export interface BroccoliConcatOptions {
...
+   * List of classes that are used by application CSS and might conflict
+   * with the optimizer. You should add any short class names (~5 characters)
+   * to this list so the optimizer doesn't use these when building the
+   * CSS Blocks compiled output.
+   *
+   * This is a convenience alias for
+   * optimization.rewriteIdents.omitIdents.class[]. It has no effect if
+   * optimization is disabled.
+   */
+  appClasses?: string[];
```
chriseppstein, line 87: Please add this option to the `@css-blocks/ember-app` README.

packages/@css-blocks/ember-utils/src/options.ts
```diff
@@ -49,19 +49,42 @@ export interface BroccoliConcatOptions {
...
   optimization?: Partial<OptiCSSOptions>;
   /**
    * Options that control the behavior of broccoli-concat, which is used
    * to concatenate CSS files together by ember-app during postprocess.
+   *
    * If this is set to false, broccoli-concat will *not* run.
    * You'll need to add additional processing to add the CSS Blocks
    * compiled content to your final CSS build artifact.
    */
   broccoliConcat?: BroccoliConcatOptions | false;
```
chriseppstein, line 76: Can you add this option to the README as well?

chriseppstein (approved):
Looks great. I left some minor comments, going to approve the PR. Please address the comments if possible before landing.

packages/@css-blocks/ember-app/src/broccoli-plugin.ts
timlindvall: I'm adding this to the log message, but the best I can add for the time being is the CSS file where the class was found _in the postprocess tree._ Not sure how much value that will add since we don't have info on what source file the class was introduced in.

PR #499, opened 2020-09-12
Test support for Ember Apps

### LINKED ISSUE #480: CSS Blocks Test Helpers for Ember
We need to provide test helpers that allow a developer to perform assertions about what css block styles are currently applied to an element.

For debugging purposes, I [already wrote code](https://github.com/linkedin/css-blocks/blob/ce06f744dfafb16e083a8a4586128d8cee53da9f/packages/%40css-blocks/ember-app/runtime/app/services/css-blocks.ts#L56) to generate a human-readable css-blocks name for the block styles that are currently applied to an element including those that are inferred through composition and inheritance. Unfortunately, the downside of the human-readable name I have right now is that the block only known as a guid (E.g. `a4f31:scope[enabled]`) and that guid isn't stable across machines so a direct reference to the guid in test code isn't feasible.

The basic idea for testing is to put these human readable strings into the class list of the element but only when the tests are running so that the element has something like `class="a4f31:scope a4f31:scope[enabled] a0 b7 c3"`. These human readable classes won't match any selectors but they will provide a way to assert that the element has a particular block style via `element.classList.contains()`

So the first challenge is to provide a way to map some sort of consistent reference to a block from the test code and to do this without further bloating the runtime data that we're currently sending to the browser. This means that we'll need to get some metadata into the test bundle so that the test code can import it. For example:

```js
import cssBlocksTestSupport from "my-app/css-blocks-test-support";
// I'm not sure how to get a handle to the css blocks service from a test.
// it should be able to be dependency injected.
cssBlocksService.enableDebugMode(cssBlocksTestSupport);
let darknessBlock = cssBlocksTestSupport.getBlock("@linkedin/hue-web-foundations", "darkness");
let darkness = darknessBlock.style(":scope");
assert(element.classList.contains(darkness));
```

The reason we should pass `cssBlocksTestSupport` into the service call is that, in the future, it is my plan that the actual block style names won't be part of the runtime deliverable -- I should be able to optimize them away to just being a unique number. So I'd like to provide appropriate API surface that we can move the style name generation to the test meta data in the future. That call might do nothing with the test metadata right now, but we should require it in order to put css blocks into debug mode so that we can use it in the future.

We should be able to generate a javascript file that contains the API and the relevant data using the [`treeForTestSupport`](https://ember-cli.com/api/classes/Addon.html#method_treeForTestSupport
) hook from our `@css-blocks/ember-app` addon.

The tree containing the compiled block files in the final application build is probably the easiest way to construct human meaningful paths to the blocks. In that tree there's a subdirectory for the application, a subdirectory for addons, and a subdirectory for lazy engines. [This is some code](https://github.com/linkedin/css-blocks/blob/ce06f744dfafb16e083a8a4586128d8cee53da9f/packages/%40css-blocks/ember-utils/src/BroccoliTreeImporter.ts#L43-L52) where I search through the various subdirectories to look for references to block files.

ramitha:
Resolves #480 

Adding test support for ember apps that depend on @css-blocks/ember-app. 

We introduce a test-support method called `setupCSSBlocksTest` that's responsible for setting up the test infrastructure.
- This involves extending the `-css-blocks` service and overriding the class names omitted by the service to include fake classNames for all the elements. 
- This method also exposes the CSSBlocksTestService itself via `this.cssBlocks` that has a single API called `getBlock` that can then be queried for these fake classNames. 

For this API to function, we generate `css-blocks-test-support-data` during build that the testService can talk to.


packages/@css-blocks/ember-app/src/brocolli-plugin.ts
```diff
@@ -63,9 +66,26 @@ export class CSSBlocksApplicationPlugin extends Filter {
...
       let filename = importer.debugIdentifier(block.identifier, config);
+
+      if (!this.isProdApp) {
+        // Generate the test data
+        // Iterate over the blocks that belong to this app and its addons only for
+        // generating the test support data
+        if (block.identifier.startsWith(IDENTIFIER_PREFIX)) {
+          let filePath = pathToIdent(filename).split(".compiledblock.")[0].replace(IDENTIFIER_PREFIX, "");
+          // locate the actual block corresponding to this compiled block to
+          let factoryBlock = await factory.getBlock(block.identifier);
```
chriseppstein, line 82: I don't understand why you're using the factory to get the block again. I think you can just do `block.eachBlockExport()`.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
```diff
@@ -63,9 +66,26 @@ export class CSSBlocksApplicationPlugin extends Filter {
...
     for (let block of blocksUsed) {
+
       let content: postcss.Result;
       let filename = importer.debugIdentifier(block.identifier, config);
+
+      if (!this.isProdApp) {
+        // Generate the test data
+        // Iterate over the blocks that belong to this app and its addons only for
+        // generating the test support data
+        if (block.identifier.startsWith(IDENTIFIER_PREFIX)) {
```
chriseppstein, line 79: There's a method to check for this in `BroccoliTreeImporter`:

```suggestion
        if (isBroccoliTreeIdentifier(block.identifier)) {
```

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
```diff
@@ -63,9 +66,26 @@ export class CSSBlocksApplicationPlugin extends Filter {
...
+          let filePath = pathToIdent(filename).split(".compiledblock.")[0].replace(IDENTIFIER_PREFIX, "");
```
chriseppstein, line 80: I don't understand why you're working your way from the filename to the identifier with `pathToIdent()` when you can just use `block.identifier`.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
```diff
@@ -63,9 +66,26 @@ export class CSSBlocksApplicationPlugin extends Filter {
     }
     let compiler = new BlockCompiler(postcss, config);
     let reservedClassnames = analyzer.reservedClassNames();
+    let testDataBuilder = new TestSupportDataGenerator();
     for (let block of blocksUsed) {
+
```
chriseppstein, line 71: This is going to grab every block that's in the broccoli tree which is every addon, every engine, and the app too. I think this exposes too much to the tests. I don't think the tests should be able to access any block that isn't one of their app's blocks or exported from one of their app's own blocks. Can we restrict this to just the application's blocks? If not, what is the use case that I'm missing here?

chriseppstein (changes requested):
This is a great start @ramitha! I really like the overall structure of the testing solution that you've put together here. I think most, if not all, of the issues I've raised here are going to be fairly straightforward to resolve.

packages/@css-blocks/ember-app/runtime/app/services/css-blocks.ts
```diff
@@ -29,25 +27,45 @@ interface StyleIdToOptimizationsMap {
 export default class CSSBlocksService extends Service {
   possibleOptimizations!: StyleIdToOptimizationsMap;
   styleNames: { [name: string]: string };
+  static enableDebugMode: Boolean = false;
+  static enableTestMode: Boolean = false;
```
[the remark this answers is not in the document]
ramitha: Ah, thank you!

packages/@css-blocks/ember-app/runtime/app/services/css-blocks.ts
```diff
@@ -29,25 +27,45 @@ interface StyleIdToOptimizationsMap {
...
     this.styleNames = getStyleNames();
   }
   classNamesFor(argv: Array<string | number | boolean | null>): string {
-    if (DEBUGGING) {
+    if (CSSBlocksService.enableDebugMode) {
       console.log(argv);
     }
+    let stylesApplied = this.getDirectlyAppliedStyles(argv);
+
+    return this.getImpliedAndOptimizedStyles(stylesApplied);
```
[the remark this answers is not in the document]
ramitha: Sure. I refactored this bit initially when I was using this method to get the actual classNames.

packages/@css-blocks/ember-app/runtime/app/services/css-blocks-test-support.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+ */
+// TODO: get qunit types
+/// @ts-ignore
+export function setupCSSBlocksTest(hooks) {
+  hooks.beforeEach(function() {
+    /// @ts-ignore
+    if (!this.owner) {
+      throw new Error(
+        "setupTracking must be called after setupTest|setupRenderingTest|setupApplicationTest",
```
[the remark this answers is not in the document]
ramitha: Yes, I used ember-cli-pemberly-tracking as a reference for how to provide test scaffolding.

packages/@css-blocks/ember-app/runtime/app/services/css-blocks-test-support.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      test('visiting /', async function (assert) {
+        await visit('/');
+        let defaultBlock = this.cssBlocks.getBlock("hue-web-component/styles/components/hue-web-component", "default");
+        let element = find('[data-test-large-hello]');
+        assert.ok(element.classList.contains(defaultBlock.style(':scope[size="large]')));
+      });
+    });
+ */
+// TODO: get qunit types
```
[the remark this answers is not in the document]
ramitha: Good point. The standard way is to use `ember-qunit`, but people can also use `ember-mocha` and both of these provide the same functionality for setting up different types of ember tests. I'll make a note of it in the comments.

packages/@css-blocks/ember-app/runtime/app/services/css-blocks-test-support.ts
```diff
@@ -0,0 +1,98 @@
...
+const testData: TestSupportData = _testData;
+
+export class TestBlock {
+  blockId: string;
+  constructor(blockId: string) {
+    this.blockId = blockId;
+  }
+  style(styleName: string) {
+    // we need to strip "" for styles like [size="large"]
+    return `${this.blockId}${styleName.replace('"', "")}`;
```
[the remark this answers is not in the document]
ramitha: I'm also debating whether we want to provide a more meaningful error message here with the list of available styles if the style supplied is not found

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
ramitha: Good catch. This was before the export bug was filled and when I was trying to see if the original block contained the exported blocks.

ramitha: Hmm.. the block in context is the compiled block that doesn't have the exported blocks on it. Could this be another bug?

packages/@css-blocks/ember-app/runtime/app/services/css-blocks-test-support.ts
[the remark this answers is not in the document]
chriseppstein: @ramitha That's a great idea!

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
ramitha: I'm iterating through all the blocks because I not only need the blocks exposed by the app, but also those of an addon. The use case here is, if I have an ember addon that contains a component, the dummy app should be able to access the blocks for the component in its acceptance tests. 

The way I'm filtering down to the app's blocks and the addon's blocks while building the test data is by checking if the `block.identifier` starts with `IDENTIFIER_PREFIX`(line 79). If it's an absolute path, we can assume that it's not a part of this module and we don't add that to the test data.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
chriseppstein: Yes, the compiled block should have the exported blocks. If not there, that's definitely a bug.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
chriseppstein: > The use case here is, if I have an ember addon that contains a component, the dummy app should be able to access the blocks for the component in its acceptance tests.

ok. So the specific use case is that the tests want to test the addon's internals without exposing the addon's blocks via the `css-blocks.main` property in `package.json`? That's a good use case. Thanks.

The current implementation appears to expose all addons to the tests, even for applications that are not addons undergoing test.

I think we can handle this specific situation by using block aliases to expose the addon's blocks to the dummy app's templates without exposing the blocks to NPM import. The export doesn't have to be used in the template, it can just exist for testing purposes. If we do that, we can simplify the metadata generation to just blocks that are part of the main application, yes?

packages/@css-blocks/ember-app/runtime/app/services/css-blocks-test-support.ts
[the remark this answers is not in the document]
chriseppstein: > both of these provide the same functionality for setting up different types of ember tests.

Does this mean they provide compatible APIs for our needs and that our code will work for both?

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
chriseppstein: Does the fix to `@css-blocks/core/src/BlockParser/features/export-blocks.ts` not remedy this?

packages/@css-blocks/ember-app/runtime/app/services/css-blocks-test-support.ts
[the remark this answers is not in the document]
ramitha: Yes, they do. `ember-mocha` also exposes the same setup functions and sets `this.owner` (from their README) and `mocha` has a `beforeEach` hook,  so our code will work. It looks like we aren't using it at all at LinkedIn though, so I haven't tested it.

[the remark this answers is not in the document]
ramitha: Also, any framework can override the runtime service to use the test service instead (if they have to using different hooks)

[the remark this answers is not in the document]
chriseppstein: Ok. As long as the main use cases are covered here, I'm happy. We can tackle other testing frameworks on an as-needed basis.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
ramitha: No, it doesn't. I better write a test for this.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
ramitha: As discussed, this will expose blocks within in-repo addons and in-repo engines to the tests as well. Anything that is not in-repo will not be exposed to the test. Resolving this for now.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
ramitha: Sorry for the mess. Cleaned it up a bit.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
ramitha: Turns out the test passes but I was missing a yarn link into @css-blocks/core.

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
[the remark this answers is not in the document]
ramitha: Done

packages/@css-blocks/ember-app/src/brocolli-plugin.ts
chriseppstein: That's great news!

packages/@css-blocks/ember-app/runtime/app/services/css-blocks.ts
```diff
@@ -29,25 +27,45 @@ interface StyleIdToOptimizationsMap {
...
+   * @param stylesApplied set of styleIds that are directly applied on the element
+   */
+  getImpliedAndOptimizedStyles(stylesApplied: Set<number>): string {
     let resolver = new StyleResolver(data);
     for (let style of stylesApplied) {
       resolver.addStyle(style);
     }
 
-    if (DEBUGGING) {
+    if (CSSBlocksService.enableDebugMode) {
```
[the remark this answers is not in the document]
ramitha: I really liked what you suggested about debugging in prod and I was curious to see how to do this. If we have the ember inspector enabled, I can peek into the service instance and its values. So I've made a change to add this on the instance itself.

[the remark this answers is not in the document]
chriseppstein: @ramitha oh cool. I'd love to see a demo of that!

